PR #37, opened 2021-11-10
docs: replace hardcoded strings

### LINKED ISSUE #10: Replace hardcoded strings with translation method in *.tsx files
Prepare the static docsite strings for translation:
- [ ] Replace hardcoded strings with the `t('***')` method in all `*.txs` files
   *Check out `pages/team.tsx` line `115` to see the translation method in use*
- [ ] Add those strings in `i18n/ui.json`

wattanx:
Closes #10 

## 📝 Description

Replace hardcoded strings with the t('***') method in all *.txs files

## 🚀 New behavior

The hard-coded string will be displayed with the value based on ui.json.

No

## 📝 Additional Information
Sentences such as the following may be difficult to translate.
image


lazarnikolov94 (changes requested):
LGTM! Only thing missing that I've missed when replacing the hardcoded strings is the "homepage.title" string. The "with speed" part of the string should be in a different color. Think about how we can achieve translation + coloring that part with teal color. Have in mind that in some languages the "with speed" part might not be at the end of the sentence. See if there's any formatting capabilities in i18n that we can use.

wattanx:
I've also replaced the "homepage.title" string.
By using Trans Component, we can handle cases where "with speed" is not at the end of the sentence.
https://react.i18next.com/latest/trans-component

lazarnikolov94:
> I've also replaced the "homepage.title" string. By using Trans Component, we can handle cases where "with speed" is not at the end of the sentence. https://react.i18next.com/latest/trans-component

That's exactly what we want! Let's just resolve the merge conflicts, and then we can close this issue. Thanks @wattanx!

lazarnikolov94 (approved):
That's it! LGTM!

lazarnikolov94:
@allcontributors please add @wattanx for documentation.

PR #140, opened 2021-12-16
doc: mention color "mode" function in the doc

giacomorebonato:
# Possible doc improvement

## 📝 Description

I had difficulties to find a way to define in the theme a color that changes depending on the color mode.
This is handy so I don't need to handle it inside of each component (with `useColorModeValue`).
I don't know if I am missing something else from the documentation, but I wish I would have found sooner the mention of the function `mode` from `@chakra-ui/theme-tools`.


lazarnikolov94:
Thanks @giacomorebonato!

@allcontributors please add @giacomorebonato for documentation.

PR #155, opened 2021-12-18
Revert "doc: mention color "mode" function in the doc"

giacomorebonato:
I am sorry, but this is not working as expected.
The `mode()` utility function is not meant to be used in the color section, because it expects only static values. 
I'll come back with a better contribution and example, sorry about that.

Reverts chakra-ui/chakra-ui-docs#140


lazarnikolov94:
Ah, right. To use the `mode` you'd need to pass props.

PR #174, opened 2021-12-25
fix: Use GridItem instead of Box

EmmanuelDemey:
We do not import Box at the beginning of this page. So I think, we should use GridItem instead.

Closes # 

## 📝 Description

Use GridItem composent instead of Box

## 🚀 New behavior

No


nikolovlazar (approved):
You're right. It doesn't change anything, but it makes it clearer. Thanks @EmmanuelDemey.

nikolovlazar:
@allcontributors please add @EmmanuelDemey for documentation.

PR #194, opened 2022-01-02
Add notes in `Usage` section

tannguyencse19:
See this discussion: https://github.com/chakra-ui/chakra-ui/issues/3018#issuecomment-1003737292

### DIFF
--- a/pages/docs/hooks/use-breakpoint-value.mdx
+++ b/pages/docs/hooks/use-breakpoint-value.mdx
@@ -27,7 +27,7 @@ The `useBreakpointValue` hook returns the value for the current breakpoint.
 
 ## Usage
 
-> `useBreakpointValue` need `base` value (not `sm`) to not return `undefined` in first render.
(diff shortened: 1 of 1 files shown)


pages/docs/hooks/use-breakpoint-value.mdx
```diff
@@ -27,6 +27,8 @@ The `useBreakpointValue` hook returns the value for the current breakpoint.
 
 ## Usage
 
+> `useBreakpointValue` need `base` value (not `sm`) to not return `undefined` in first render.
```
nikolovlazar, line 30: > Make sure to provide a `base` value when using `useBreakpointValue` so it doesn't return `undefined` in the first render.

nikolovlazar (changes requested):

pages/docs/hooks/use-breakpoint-value.mdx
tannguyencse19: Yeah, use your way.

PR #214, opened 2022-01-10
fix(docs): Fix pagination order in hooks docs

rodrigofuentes:
## 📝 Description

Fixes the order of Pagination for the hooks docs. Specifically the order of `useConst` and its previous and next order in its sibling pages.

There is a disorder of navigation on the pages and skips back an extra page on the siblings of `useConst`/`useControllable`

## 🚀 New behavior

Pagination works as expected.

No

### DIFF
--- a/configs/docs-sidebar.json
+++ b/configs/docs-sidebar.json
@@ -422,12 +422,12 @@
               "path": "/docs/hooks/use-clipboard"
             },
             {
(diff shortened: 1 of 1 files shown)


nikolovlazar:
Makes sense!

@allcontributors please add @rodrigofuentes for bugs.

PR #215, opened 2022-01-10
Clarify the useClipboard timeout parameter

bintay:
Closes # 

## 📝 Description

The useClipboard documentation wasn't super clear, so I added information about defaults & units. I'm not particular about the language used, but it would be nice to have the default timeout of 1500ms specified!

The `optionsOrTimeout` parameter is confusing.

## 🚀 New behavior

Adds more details about `optionsOrTimeout`.

No

## 📝 Additional Information

Perhaps it would be even better if `optionsOrTimeout` was replaced with two parameters in the code, `options` and `timeout`. Then each idea is its own discrete unit that can be understood without reference to the other, reducing mental load.

### DIFF
--- a/pages/docs/hooks/use-clipboard.mdx
+++ b/pages/docs/hooks/use-clipboard.mdx
@@ -11,20 +11,20 @@ description:
 
 The `useClipboard` hook takes the following arguments:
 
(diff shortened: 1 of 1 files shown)


nikolovlazar:
@allcontributors please add @bintay for docs.

PR #236, opened 2022-01-19
docs: put showcase section in homepage

### LINKED ISSUE #223: Add showcase to the homepage
### Subject

Homepage

### Description

Add a showcase section to the homepage taking some interaction ideas from here: https://library.relume.io/

https://cln.sh/gNLUzS

nikolovlazar:
Closes #223 

## 📝 Description

Adding a "Showcase" section in the homepage.

No

## 📝 Additional Information

The section has a fine interactivity on Desktop and other devices that have cursors. On `md` and down and "touch-only' devices the interactivity is disabled, and the showcase items are displayed in a simple way.

## 👀 Preview

https://user-images.githubusercontent.com/5396211/150210810-63917d52-73ec-4ba9-ad21-cdea68a1e253.mp4


src/components/showcase/showcase-section.tsx
```diff
@@ -0,0 +1,213 @@
...
+              px='40px'
+              fontSize='1.2rem'
+              as='a'
+              size='lg'
+              colorScheme='teal'
+              rightIcon={<FaArrowRight fontSize='0.8em' />}
+            >
+              {t('homepage.built-with-chakra-section.see-showcase')}
+            </Button>
+          </NextLink>
```
noobinthisgame, line 188: If you use Button with as="a" you could also just do Button as={NextLink} or am I missing something?

nikolovlazar: All of these are external links, so we don't need to use client-side navigation. That's why I used a simple anchor tag instead of `NextLink`.

noobinthisgame: But here it points to our "/showcase" page ?

noobinthisgame: I mean that link in particular. The actual showcase elements I agree

nikolovlazar: The issue with that is that the `Button` component renders multiple children (label, arrow icon), but `NextLink` only supports a single child:
`Multiple children were passed to <Link> with `href` of `/showcase` but only one child is supported`

noobinthisgame: Alright then it's fine for me :)

PR #275, opened 2022-02-02
Update gatsby-guide.mdx

### LINKED ISSUE #274: @chakra-ui/gatsby-plugin docs isn't related to Gatsby
### Description

In the guide here https://chakra-ui.com/guides/getting-started/gatsby-guide, the section containing information on extending the default theme isn't clear. 
In gatsby, there isn't an `<App />` component but rather `element` exposed through the `wrapRootElement` API in gatsby-browser.js. Hence, the demo relates to general React use cases instead of gatsby.js. 

I'm presently finding it difficult to add custom themes as different texts point out different ways to do this, that don't seem to work.

nikolovlazar:
Closes #274 

## 📝 Description

The Gatsby guide is not clear enough. I found another Gatsby guide under `pages/guides/integrations/with-gatsby.mdx` which has the correct way of using Chakra UI with Gatsby projects. I'm moving over the contents from that file into the `getting-started/gatsby-guide.mdx` file.

The Gatsby guide contains unclear instructions on how to setup Chakra UI in Gatsby projects.

## 🚀 New behavior

The Gatsby guide contains the correct instructions on how to setup Chakra UI in Gatsby projects.

No


nikolovlazar:
@Chuloo could you take a look at this and let me know if the instructions are clear and correct?

Chuloo:
Hi @nikolovlazar yes this looks good, however, if I don't want to shadow the plugin to customize the theme, chakra-ui has an `extendTheme` API if I understand correctly. Can you include how this is used in Gatsby applications? I assume since ChakraProvider is already wrapping the root element by default, is it possible to wrap the same element again only this time, passing the extended theme object. Do you know how this would look?

nikolovlazar:
> Hi @nikolovlazar yes this looks good, however, if I don't want to shadow the plugin to customize the theme, chakra-ui has an `extendTheme` API if I understand correctly. Can you include how this is used in Gatsby applications? I assume since ChakraProvider is already wrapping the root element by default, is it possible to wrap the same element again only this time, passing the extended theme object. Do you know how this would look? 

I don't think you can customize the theme without shadowing the plugin. As you can see, the wrapper element inly takes the page, and a couple of props. The theme is only referenced from the plugin's theme file.

Chuloo:
Okay, would be nice to do that in Gatsby-browser tho. Thanks for the help, I'll try to shadow the plugin and extend the theme to see if that works.

Chuloo:
Customizing the chakra theme by shadowing still doesn't work. Are you able to reproduce this too? I followed the sample outlined here https://github.com/chakra-ui/chakra-ui/tree/main/tooling/gatsby-plugin, yet it doesn't work. Could it be related to this? https://github.com/chakra-ui/chakra-ui/pull/362 for the old plugin.

PR #284, opened 2022-02-05
docs: fix missing children in theming example

aaarichter:
## 📝 Description

Adds `{children}`to theming example @ https://chakra-ui.com/docs/theming/component-style#return-value

Theming example destructs `children` in line 222 but doesn't return it

## 🚀 New behavior

removes destructed `children` from theming example props

No

### DIFF
--- a/pages/docs/theming/component-style.mdx
+++ b/pages/docs/theming/component-style.mdx
@@ -224,7 +224,7 @@ function Card(props) {
   const styles = useStyleConfig('Card', { variant })
 
   // Pass the computed styles into the `__css` prop
-  return <Box __css={styles} {...rest} />
+  return <Box __css={styles} {...rest} >{children}</Box>
 }
 ```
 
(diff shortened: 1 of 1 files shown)


nikolovlazar:
Good catch, @aaarichter! Although, I'd prefer if we remove the `children` destructuring completely, so it becomes a part of the `...rest` spread. Since we're not doing anything specific to the `children`, destructuring it and manually rendering it seems redundant. So let's remove it. What do you think?

nikolovlazar (changes requested):
Good catch, @aaarichter! Although, I'd prefer if we remove the `children` destructuring completely, so it becomes a part of the `...rest` spread. Since we're not doing anything specific to the `children`, destructuring it and manually rendering it seems redundant. So let's remove it. What do you think?

aaarichter:
@nikolovlazar feedback addressed. 

I wasn't sure which way to write it. Destructing `children` in this example is redundant but from a understanding point of view a bit clearer (to me)

nikolovlazar:
@allcontributors please add @aaarichter for bugs.

PR #263, opened 2022-01-30
docs: Add Filters Section to Style-Props Page

sohamsshah:
Closes #258 

## 📝 Description

This PR adds Filter Section to the Docs by taking reference from the prop definition defined in [`packages/styled-system/src/config/filter.ts](https://github.com/chakra-ui/chakra-ui/blob/main/packages/styled-system/src/config/filter.ts)

Have added examples for:
- Filter (with and without filter="auto")
- Backdrop Filter (with and without backdrop-filter="auto")

## ⛳️ Current behavior (updates)

Currently, the section doesn't exist in Styled-Prop definition.

## 🚀 New behavior

This PR adds the below documentation:
image
image
image

## 💣 Is this a breaking change (Yes/No):

No

## 📝 Additional Information
Will be active to hear if any changes or suggestions are required in the provided examples or the Prop Definitions. Thank you!

PR #299, opened 2022-02-10
docs: add portal hints for overlay components

m4x3d:
Closes -> no current issue

## 📝 Description

When you try to blur the background on opening e.g. a modal it will not work because the modal will be rendered in a portal. The only thing that can be blurred from this point is the modal itself or the children, which is obviously not the idea behind it.

No direct hint that the overlay components are rendered within a Portal (only via props).

## 🚀 New behavior

Direct hint that the overlay components are rendered within a Portal.

No

## 📝 Additional Information
To be 100% correct the information needs also be on the Tooltip Page. But I hope that nobody come to the idea to try this. :laughing: 

I resolved this at work by adding a BlurContext which contains only a bool state hook so that I can use the bool in my pagelayout component. I am not happy with that solution and open for suggestions. 
I also considered the docs of https://github.com/chakra-ui/chakra-ui-docs/pull/263.

### DIFF
--- a/pages/docs/overlay/alert-dialog.mdx
+++ b/pages/docs/overlay/alert-dialog.mdx
@@ -8,7 +8,9 @@ description:
 ---
 
 `AlertDialog` component is used to interrupt the user with a mandatory
-confirmation or action.
+confirmation or action. 
+It is rendered within a `portal` which mean u can not set e.g. a `filter` 
+with blur for the rest of your application.
 
 <ComponentLinks
   github={{ package: 'modal' }}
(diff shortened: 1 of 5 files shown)


nikolovlazar:
Actually @m4x3d, you can add blur behind a modal: https://codesandbox.io/s/modal-backdrop-blur-eoy2r

nikolovlazar:
Yeah you could add a new section in the Modal page "Styling the backdrop" and explain some ways on how to style the backdrop, either with a blur, or change the opacity/color etc... Hack around, see what you can figure out, and package everything in a PR.

m4x3d:
Figured out that it works on Chrome and all the other common browsers except Firefox (and IE, lol) -> https://caniuse.com/?search=backdrop-filter

So I guess I dont be so dumb as I felt yesterday :rofl: 

Try to figure out if if there is a possibility to polyfill this behavior on Firefox because I dont think the footnote on caniuse regarding some FF settings is a solution (every user needs to that by itself?!).

If you have any idea let me know!

nikolovlazar:
@m4x3d that's true. I guess a workaround would be to use an SVG filter instead of the backdrop filter property.

Issue #313, opened 2022-02-12
Git clone fails on windows.


wattanx:
## Description
Git clone fails on windows.

## Error Message
The following message will be output when you clone the file.

> error: invalid path 'public/showcases/articles/Choosing-the-right-component-library-for-your-design-system:-MUI-vs-Chakra.png'
fetal: unable to checkout working tree.
warning: Clone succeeded, but checkout failed.

Probably the reason is that `:` is not available in the Windows file system.

## Operating System
Windows 10

PR #316, opened 2022-02-13
fix: remove semicolon

wattanx:
Closes # https://github.com/chakra-ui/chakra-ui-docs/issues/313

## 📝 Description
Git clone fails on windows.

The reason is that `:` is not available in the Windows file system.

showcases/articles/Choosing-the-right-component-library-for-your-design-system:-MUI-vs-Chakra.png

## 🚀 New behavior
remove semicolon
showcases/articles/Choosing-the-right-component-library-for-your-design-system-MUI-vs-Chakra.png

No


nikolovlazar:
Good one @wattanx! We should also fix the way we generate the showcase items. Would you like to take a look at that?

wattanx:
Thanks! 
I'd like to try it.

nikolovlazar:
Awesome! I've created a new issue #317 on which you can work on. Just write a comment so I can assign you to it, @wattanx.

PR #331, opened 2022-02-16
docs: add section for state depending behavior of switch component

### LINKED ISSUE #161: Switch component does not apply an `isInvalid` style
### Description

When the `isInvalid` prop is applied to the `Switch` component. I would expect some visual representation of this.

### Link to Reproduction

https://codesandbox.io/s/chakra-switch-isinvalid-efj7b?file=/src/index.tsx

### Steps to reproduce

1. Go to `index.tsx`
2. See that the component with `isInvalid` has no styling applied

### Chakra UI Version

1.6.12

### Browser

Google Chrome 95

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

The [docs](https://chakra-ui.com/docs/form/switch#props) state that it updates the state of the unchecked state. Not sure if the component or the documentation needs to change.

m4x3d:
Closes #161 

## 📝 Description

Add a section for boolean state prop depending behavior and styles for the Switch component, including a hint for not changing styles regarding e.g. the `isInvalid` prop.

No examples for the `Switch` component with boolean state props, like e.g. `isDisabled`.

## 🚀 New behavior

New section with examples for all boolean state props of the `Switch` component.

No


nikolovlazar (changes requested):
Looking good, @m4x3d! But, I noticed that there are responsiveness issues in the `FormControl`:

CleanShot 2022-02-19 at 21 36 28@2x

m4x3d:
> Looking good, @m4x3d! But, I noticed that there are responsiveness issues in the `FormControl`:
> 
> CleanShot 2022-02-19 at 21 36 28@2x

Fixed it. @nikolovlazar 

Before breakpoint is reached:
image

Base, directly below breakpoint:
image

pages/docs/form/switch.mdx
```diff
@@ -62,6 +62,33 @@ You can change the checked background color of the `Switch` by passing the
 </Stack>
 ```
 
+## State depending behavior
+
+States like `isDisabled` have an impact on the usability of a `Switch` and on
+the styles, except for the `isInvalid` and the `isRequired` prop.
+
+```jsx
+<FormControl as={SimpleGrid} columns={[2, null, null, 4]}>
```
nikolovlazar, line 71: I think it's better if you used the object syntax instead of the array syntax, so you'd avoid adding `null`s.

nikolovlazar (changes requested):

pages/docs/form/switch.mdx
m4x3d: Yes, you are right, it is better readable. Force-pushed it.

PR #345, opened 2022-02-21
Remove and ignore .DS_Store files

selbekk:
## 📝 Description

This pull request removes existing .DS_Store files and add them to .gitignore for ignoring future files.

.DS_Store files are just garbage files from MacOS, and we don't need them in our repo!


.gitignore
```diff
@@ -15,6 +15,7 @@ yarn-error.log
 tsconfig.tsbuildinfo
 .next
 .idea
+.DS_Store
```
m4x3d, line 18: This entry exists already in this file, see line 86.

m4x3d (commented):
Hey @selbekk, thanks for cleaning up that unnecessary DS_Store files! :heart: 

The .DS_Store files are already in the gitignore but possible someone does not remove the existing files when adding this to the gitignore.

selbekk:
There we go, should be good now. Thanks for spotting my error @m4x3d :)

nikolovlazar:
Thanks for this PR @selbekk!

PR #323, opened 2022-02-14
docs: add overview pages for components and styled-system

noobinthisgame:
## 📝 Description

This PR adds an overview page for the new components and styled-system section to show a preview of what can be found in the section.

## 💣 Is this a breaking change (Yes/No):

No

## 📝 Additional Information
This PR is based on the work the following pr https://github.com/chakra-ui/chakra-ui-docs/pull/301

PR #301, opened 2022-02-10
feat: rework docs structure and improve discoverability

noobinthisgame:
## 📝 Description

This PR reworks the structure of the docs to improve the discoverability of content and improve the UX. Also this rework aims to make it easier for new chakra user to get into chakra.

## 💣 Is this a breaking change (Yes/No):

Yes, if people might have used links somewhere to reference pages in the docs, those pages might have moved

PR #353, opened 2022-02-24
docs(structure): rework docs structure

nikolovlazar:
## 📝 Description

The main feature branch for the docs structure rework. Previously two PRs were merged into this branch: #301 and #323.

## 🚀 New behavior

The docsite's content is restructured. The content is split into two categories: **Styled System** and **Components**. Two new pages were introduced: **Styled System Overview** and **Components Overview**. All URLs are changed accordingly, and proper redirects are put in place to ensure links shared on social media still work. All work done by @noobinthisgame ⭐️

No


jacobhq (commented):
Maybe put the hooks into a separate section, or to the components section

noobinthisgame:
Thanks @jacobhq for the feedback. As the hooks are beeing used under the hood to provide logic and theming to the components I don't think they should be placed in the components section. This one just regroups ready to use components. Further the hooks are part of the Chakra UI API which leads me to think they are well kept in the styled-system section that regroups feature and API documentation. Maybe in the future once there is more hooks it can make sense to split them up in their own section.

jacobhq:
That makes sense, thank you!

PR #364, opened 2022-02-28
docs(examples): Add custom input and controls examples

jacobhq:
## 📝 Description

I've added a new example with a custom input, custom controls and a UI.

No

### DIFF
--- a/pages/docs/form/editable.mdx
+++ b/pages/docs/form/editable.mdx
@@ -46,7 +46,7 @@ import { Editable, EditableInput, EditablePreview } from '@chakra-ui/react'
 </Editable>
 ```
 
-### With custom controls
+### With custom input and controls
 
 In some cases, you might need to use custom controls to toggle the edit and read mode. We use the render prop pattern to provide access to the internal state of the component. You may want to customize the `EditableInput` as well. This can be achieved by using a custom `Input` component in the `as` prop.
 
--- a/pages/docs/form/editable.mdx
+++ b/pages/docs/form/editable.mdx
@@ -90,6 +90,54 @@ function CustomControlsExample() {
 }
 ```
 
+### With custom input and controls
+
+Building on the last example, you may want to customise the input as well. I've also tweaked the UI, by adding a hover color, and a tooltip.
+
+```jsx
+function App() {
+  /* Here's a custom control */
+  function EditableControls() {
+    const {
+      isEditing,
+      getSubmitButtonProps,
+      getCancelButtonProps,
+      getEditButtonProps
+    } = useEditableControls();
+
+    return isEditing ? (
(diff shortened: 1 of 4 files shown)


pages/docs/form/editable.mdx
```diff
@@ -90,6 +90,54 @@ function CustomControlsExample() {
 }
 ```
 
+### With custom input and controls
+
+Building on the last example, you may want to customise the input as well. I've also tweaked the UI, by adding a hover color, and a tooltip. As you can see, we can change the input by setting the `as` prop. This design is inspired by the [Atlassian inline edit](https://atlassian.design/components/inline-edit/examples) component.
```
noobinthisgame, line 95: I would rephrase this like this: 
You may want to customize the `EditableInput` as well. This can be achieved by using a custom `Input` component in the `as` prop.
// example

noobinthisgame (commented):
Hey @jacobhq, thanks for opening this PR! 
I like the idea to add more details to custom EditableInputs.
I have some thoughts about this:
- I don't think it's necessary to have such a big example combining custom controls and custom EditableInput. An example showing only how to customize the Input would be sufficient imo. What do you think ? 
- Also I would not add the Tooltip to this example, it adds complexity to it without bringing much value to the actual explanation. A well styled example of an editable would be better placed as a recipe.

jacobhq:
@noobinthisgame What about merging the custom controls and custom input examples, removing the tooltip and additional styling, but putting the styled example into a recipe?

noobinthisgame:
@jacobhq If someone wants to search for custom editable input in the search I think it makes it easier to find when in separate example. But I'd be fine to merge them for now. 
Removing styles and moving them to recipes sounds great! 
Don't forget to update the search as last step (forgot before good we talked about it :D). Let me know if you need help with that.

jacobhq:
@noobinthisgame I have created the recipe and merged the examples, please could I have some help updating the search and recipe sidebar?

noobinthisgame:
Sure! What's up with the sidebar? 

To generate the search once you are done, you can use the command `yarn search-meta:gen` (see package.json). This will trigger the script that generates new search content. :)

jacobhq:
> What's up with the sidebar?

The Atlaskit editable recipe does not appear in the sidebar. Should I manually update `configs/guides-sidebar.json` to include the page @noobinthisgame?

noobinthisgame:
Yes, that's right. The `configs/guides-sidebar.json` has to be updated manually :)

noobinthisgame (approved):
Yup this now looks ready to go 🚀

nikolovlazar:
Good one, @jacobhq!

@allcontributors please add @jacobhq for docs.

PR #370, opened 2022-03-01
docs: add hint for tabs on custom radio button example

m4x3d:
Closes -> no current issue

## 📝 Description

Added a hint on the [Custom Radio Button](http://localhost:3000/docs/form/radio#custom-radio-buttons) example to use `Tabs` instead of code in the example if there is no special need for a `Radio Button`.

No hint. New users could use this example without knowing about the `Tabs` component.

## 🚀 New behavior

Hint for using `Tabs` when there is no need for a `Radio Button`.

No

## 📝 Additional Information

This is a result of a question of a user on the Chakra UIs Discord.


pages/docs/form/radio.mdx
```diff
@@ -122,6 +122,12 @@ help with this scenario. Here's what you need to do:
 1. Create a component that consumes the `useRadio` hook.
 2. Use the `useRadioGroup` hook to control a group of custom radios.
 
+> Please be aware that the example below should only be used if you really need
+> a radio button 'under the hood'. If not and you e.g. only want to show
+> different content on activation of a button it could be much easier to use the
+> `Tabs` component, e.g. with `variant='unstyled'` to get the styling of the
+> example.
```
nikolovlazar, line 129: > Please be aware that the example below should only be used if you really need
> a radio button for data collection purposes. If you want to toggle between different
> content on activation of a button use the `Tabs` component.

nikolovlazar (changes requested):
I simplified the text a bit. Otherwise it looks good!

jacobhq (approved):

PR #372, opened 2022-03-01
Fix installation step

noahlitvin:
Currently, following the instructions as written results in `no matches found: @emotion/react@^11`.


nikolovlazar:
Hi @noahlitvin! Thank you for opening this PR! I tried to reproduce the issue, but couldn't do it. On the other hand, we've seen backwards incompatibility with framer motion before, and that's when we introduced fixed versions in the installation script. Right now it will work if you try to install Chakra UI without the fixed package versions, but we might stumble upon another backwards incompatibility in future, so we decided to keep the fixed versions.

PR #378, opened 2022-03-02
fix: Grid example closing tags

jacobhq:
## 📝 Description

Currently, there is an error in the grid example, as a result of incorrect closing tags

Grid has incorrect closing tags

## 🚀 New behavior
Grid closing tags are now correct

No


nikolovlazar:
Nice one, @jacobhq! Thank you for opening this PR! ⭐️

PR #385, opened 2022-03-03
fix(style-props): add content property and example to style-props page

aaarichter:
Closes # https://github.com/chakra-ui/chakra-ui/issues/561

## 📝 Description

- add missing `content` property & example for `::before` example with extra note about quotes

- content prop is missing and no clear example for its usage

## 🚀 New behavior

- add missing `content` property & example for `::before` example with extra note about quotes

no

### DIFF
--- a/pages/docs/styled-system/features/style-props.mdx
+++ b/pages/docs/styled-system/features/style-props.mdx
@@ -577,6 +577,14 @@ import { Button } from "@chakra-ui/react"
   >
   </Box>
 </Box>
+
+// add ::before pseudo element
+<Box
+  _before={{ content: '"🙂"', display: 'inline-block', me: '5px'} // Note: the content value need an extra set of quotes!
+>
+A pseudo element
+</Box>
+
 ```
 
 | Prop                    | CSS Property                                                                                                                                                                               | Theme Field |
@@ -650,6 +658,7 @@ following props:
 | ----------------- | ------------------ | ----------- |
 | `animation`       | `animation`        | none        |
 | `appearance`      | `appearance`       | none        |
+| `content`         | `content`          | none        |
 | `transform`       | `transform`        | none        |
 | `transformOrigin` | `transform-origin` | none        |
 | `visibility`      | `visibility`       | none        |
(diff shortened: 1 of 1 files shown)


pages/docs/features/style-props.mdx
```diff
@@ -574,6 +574,14 @@ import { Button } from "@chakra-ui/react"
   >
   </Box>
 </Box>
+
+// add ::before pseudo element
+<Box
+  _before={{ content: '"🙂"', display: 'inline-block', me: '5px'} // Note: the content value need an extra set of quotes!
```
nikolovlazar, line 580: I think you're missing a closing `}` at the end.

nikolovlazar (changes requested):

pages/docs/features/style-props.mdx
nikolovlazar: Also, which prop is `me`? Were you thinking about `mr` as in `marginRight`?

aaarichter: me - marginEnd

I'll add the missing `}`

aaarichter:
@nikolovlazar feedback addressed

pages/docs/features/style-props.mdx
nikolovlazar: Also, can you move the Note comment above `_before`, so the code block doesn't scroll horizontally on desktop?

aaarichter: addressed. I force pushed it on the last commit

jacobhq (approved):
This looks good. Why is the build failing?

noobinthisgame:
@aaarichter when the build is failing in the pipeline, most of the time you can reproduce the fail locally.

nikolovlazar:
@aaarichter we just merged a big PR that restructures the docsite. You have a conflict in the `style-props.mdx` file now, because it got moved to `pages/docs/styled-system/features/style-props.mdx`.

aaarichter:
seems like the rebase did the trick. pipeline is green now. 💚 

> when the build is failing in the pipeline, most of the time you can reproduce the fail locally.

good to know. On that point I was expecting to just open the pipeline tool and see what the issue is but when faced with a login and the need to register, I bailed. I wonder if it is better than github actions, which can be viewed from anyone ?!

nikolovlazar:
@aaarichter just add an empty space after the `'5px'`, and then we merge, okay?

aaarichter:
force pushed @nikolovlazar

PR #396, opened 2022-03-04
feat: scroll sidebar when coming from overview

### LINKED ISSUE #395: Scroll sidebar when clicking a button on new "Styled System"/"Components" pages
### Description

GREAT work on the new pages yall! I love it! ❤️ 

One improvement I'd like to suggest is finding a way to scroll the sidebar to the specific section when a user clicks on one of the buttons on the new pages. 

For example, clicking "Forms" on the "Components" page redirects to the `Button` page, but the sidebar stays in place so it might not be clear to a user why they're now on the `Button` page and doesn't help the user find the other form-related components.

noobinthisgame:
Closes #395 

## 📝 Description

This adds a scroll the sidebar when the user comes from the component / styled-system overviews. Like this the user sees where he actually is on the sidebard. The scroll event is not triggered when the user uses the sidebar normally.

Sidebar is not scrolling as described in the issue. Kinda confusing for the user.

## 🚀 New behavior

Sidebar now scrolls.

No

## 📝 Additional Information
I chose to trigger the scroll event with a flag in the url as it only has to be triggered conditionally and from outside of the actual component where the event is triggered.

### DIFF
--- a/src/components/features-overview.tsx
+++ b/src/components/features-overview.tsx
@@ -106,7 +106,11 @@ export const FeaturesOverview = () => {
   return (
     <SimpleGrid mt='12' minChildWidth='15.625rem' spacing='8'>
       {features.map((feature) => (
-        <Link key={feature.title} passHref href={feature.routes[0].path}>
+        <Link
+          key={feature.title}
+          passHref
+          href={`${feature.routes[0].path}?scroll=true`}
+        >
           <Feature icon={icons[feature.title] ?? null} title={feature.title}>
             {feature.summarize
               ? `${feature.routes.length} ${changeFeatureText(feature.path)}`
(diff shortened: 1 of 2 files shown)


m4x3d:
In first instance it looks very good it!

I found a bug on the Components > Recipes link. When you click on that it creates two entries on the history, so that you have to click two times on the 'go back one page' button of your browser. It happens for me only on Firefox (97.0.1), on Chrome it works as expected.

Another point is that it could be better for the users overview to have the category name on top of the sidebar. So when you click on 'overlay' you now see the `Alert Dialog` on top of the sidebar. I could be better to have 'Overlay' instead on top. I guess this means a bit of mapping which component belongs to which category.

And it could also be good to have this feature in the other direction too. So that the scrollbar is on e.g. 'Components' after you go back in the browsers history.

noobinthisgame:
Hi @m4x3d,

Thanks for the feedback. The issue with the doubled browser history entries seem to be related to the `SandpackEmbed` component. The issue also happens on the landing page. I'd address that in a separate issue.

I pushed changes addressing the position of the scrolled element right after your comment :)

m4x3d (approved):

PR #409, opened 2022-03-09
docs(numberinput): added NumberInputField size hint.

### LINKED ISSUE #402: (docs): Improve documentation of NumberInput
### Subject

NumberInput

### Description

According to https://github.com/chakra-ui/chakra-ui/issues/5682#issuecomment-1061036448 we should update the docs of the `NumberInput` component.

nikolovlazar:
Closes #402 

## 📝 Description

Adding a short hint in the `NumberInputField` Props section about changing the size.

The current documentation says that the `NumberInputField` composes `Input` so we can pass all `Input` props. This means that if we want to change the size we can set the `size` prop to the `NumberInputField` component, which is wrong in this case. To change the size we need to pass the `size` prop to the `NumberInput` component instead.

## 🚀 New behavior

I've added an additional hint in the "NumberInputField Props" section at the bottom regarding the size.

No

### DIFF
--- a/pages/docs/components/form/number-input.mdx
+++ b/pages/docs/components/form/number-input.mdx
@@ -7,8 +7,9 @@ description:
   controls for incrementing or decrementing numeric values.
 ---
 
-The `NumberInput` component is similar to the [Input](/docs/components/form/input)
-component, but it has controls for incrementing or decrementing numeric values.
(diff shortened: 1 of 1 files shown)


noobinthisgame (approved):
Neat!

PR #416, opened 2022-03-10
Adding VStack and HStack in search-meta

### LINKED ISSUE #411: (docs): Add HStack and VStack in Search Meta
### Description

When I type `cmd+k` and type `HStack` i expect to find a link to the Stack component.

aacevski:
Closes #411 

## 📝 Description
I've added VStack and HStack as titles in the Stack page, on top of that I've updated their description and the `Adding dividers`, I've changed to Stack Dividers for a more straight forward name + ease of searching.

No.

### DIFF
--- a/configs/search-meta.json
+++ b/configs/search-meta.json
@@ -1,70 +1,70 @@
 [
   {
     "content": "Changelog",
-    "id": "7016867f-e44a-4f8d-b333-28be19d8a798",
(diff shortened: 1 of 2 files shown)


aacevski:
I'm drafting this now since when we run the `yarn search-meta:gen` it deletes it. I will add titles for VStack and HStack instead.

PR #420, opened 2022-03-12
fix(inline-code): Render `whiteSpace: nowrap` prop on component if the content does not exceed a certain width.

### LINKED ISSUE #358: Changelog long mdx code components are not responsive on mobile
### Description

On the changelog page the very long mdx code components are not responsive on mobile. This is caused by the whiteSpace property in the theme.ts (mdx.code) beeing set to nowrap. Removing this doesn't resolve the issue as it's set on purpose because code components should be in one piece. 

To solve this issue it is needed to find a way to make only the very long mdx code components responsive without breaking the current behaviour of the code components.

Screenshot 2022-02-25 at 23 53 50

TylerAPfledderer:
Closes #358 

## 📝 Description

- Remove `whiteSpace` from the extended theme config as it is now redundant.
- Add ref to the component
- Import `useDimensions` hook to retrieve the border-box width
- Create a ternary in the `whiteSpace` prop where if the border-box width is less than or equal to 363, then return the value `nowrap` else remain at the default value.

The value "363" represents the minimum width of the `li` element in the Changelog page before the page layout shifts to mobile.

The component will force no wrapping of its text no matter the parent container size.

## 🚀 New behavior

Only instances of the component that are equal to or shorter than 363px wide will receive `white-space: nowrap`. Because `whiteSpace: nowrap` value is being determined in the component, remove the instance from the extended theme file's base styles.

No


noobinthisgame (commented):
Sadly the issue with the overflowing elements is still present. (See log for the 25.02.2022)
image

TylerAPfledderer:
@noobinthisgame Hmmm. Working as expected on my end. 🤔

Is there something else I need to check/test to ensure it renders properly every time?

noobinthisgame:
@TylerAPfledderer what browser are you using ? I've checked this on chrome on the deployed preview version. Haven't done anything special tho except checking it in mobile view.

TylerAPfledderer:
I use Firefox. I've checked with Chrome and see the expected result.

When checking the Vercel preview link in Chrome, I see this:
image

noobinthisgame:
With a width of 600 it works for me as well. I tried on iphone SE and X and there it breaks again. You could try to add the wordBreak prop when no whitespace prop is set. That might help.

TylerAPfledderer:
I'll check it out! 😁

TylerAPfledderer:
@noobinthisgame I'm now viewing localhost on my Pixel 5a with Chrome and I see the issue. `wordBreak` is not immediately fixing it. I will keep exploring the problem. 👍‍

TylerAPfledderer:
@noobinthisgame so I found a strange solution.

Simply flipping the ternary operator!

So instead of
```jsx
whiteSpace={ dimensions?.borderBox.width <= MIN_CONTENT_WIDTH ? 'nowrap' : '' }
```
Now:
```jsx
whiteSpace={ dimensions?.borderBox.width > MIN_CONTENT_WIDTH ? '' : 'nowrap' }
```

I'm not quite sure why this works. I'll push this change for you to review.

TylerAPfledderer:
Thanks @noobinthisgame! I will make that update later today. 

Would you happen to know why this change worked?

noobinthisgame:
@TylerAPfledderer TBH I've tried to understand but I have no clue. I'd need to check into that deeper :/ 

It works and that's what matters :D

PR #423, opened 2022-03-13
fix(changelog): fixed changelog search meta url

nikolovlazar:
## 📝 Description

The Changelog search meta URL was pointing to `/changelog/content` because the changelog contents were kept in the `content.mdx` file. I renamed the `content.mdx` to `index.mdx` to avoid the `/content` segment when generating the search meta.

The Changelog entry in the Search is pointing to `/changelog/content`, and it goes to the 404 page.

## 🚀 New behavior

The Changelog entry now points to just `/changelog` and no 404 happens.

No

### DIFF
--- a/configs/search-meta.json
+++ b/configs/search-meta.json
@@ -1,72 +1,72 @@
 [
   {
     "content": "Changelog",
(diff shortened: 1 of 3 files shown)


noobinthisgame (approved):
Lgtm 🚀

PR #424, opened 2022-03-13
docs: use useDisclosure in alert AlertDialog example

feliskio:
## 📝 Description

Use the useDisclosure hook in first AlertDialog examle instead of useState.

The current first example for the Alert Dialog uses useState to determine whether the component should be visible or not. All other examples use the custom useDisclosure hook.

## 🚀 New behavior

The first example now uses the useDisclosure hook, too.

No


nikolovlazar:
Good catch, @feliskio! Congratulations on merging your first PR 💚

PR #425, opened 2022-03-13
docs(use-dimensions): Add doc page for the `useDimensions` hook.

TylerAPfledderer:
## 📝 Description

Build the documentation  page for the `useDimensions` hook.

## 📝 Additional Information

 The live preview does not like resolving optional chaining. Throws error `unexpected token`.
   Example: `<ListItem>The Width: {dimensions?.borderBox.width}</ListItem>`


pages/docs/styled-system/utility-hooks/use-dimensions.mdx
```diff
@@ -0,0 +1,89 @@
...
+## Import
+
+```js
+import { useDimensions } from '@chakra-ui/react'
+```
+
+## Return value
+
+This hook returns an object with the properties `marginBox`, `paddingBox`,
+`borderBox`, `conentBox`, `border`, `padding`, and `margin`.
```
nikolovlazar, line 19: `contentBox` instead of `conentBox`.

pages/docs/styled-system/utility-hooks/use-dimensions.mdx
```diff
@@ -0,0 +1,89 @@
...
+| ---------------------------- | --------------------------------------------------- |
+| `top`                        | `top`                                               |
+| `right`                      | `right`                                             |
+| `bottom`                     | `bottom`                                            |
+| `left`                       | `left`                                              |
+| `width`                      |                                                     |
+| `height`                     |                                                     |
+| `x`                          |                                                     |
+| `y`                          |                                                     |
+| `center` (x,y)               |                                                     |
```
nikolovlazar, line 34: Try flipping the table. I think the results will look better that way. Don't use columns for each property, just separate them with commas. Example:

```
Value:        | Contents:
--------------------------------------------------------------------------
marginBox    | top, right, bottom, left, width, height, x, y, center (x, y)
borderBox    | top, right, bottom, left, width, height, x, y, center (x, y)
paddingBox   | top, right, bottom, left, width, height, x, y, center (x, y)
contentBox   | top, right, bottom, left, width, height, x, y, center (x, y)
border       | top, right, bottom, left
padding      | top, right, bottom, left
margin       | top, right, bottom, left
```

This way you're clearly showing all of the `*Box` values and their contents.

pages/docs/styled-system/utility-hooks/use-dimensions.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+## Usage
+
+```jsx
+function example() {
+  const elementRef = useRef()
+  const dimensions = useDimensions(elementRef)
+
+  return (
+    <Box ref={elementRef}>
```
nikolovlazar, line 44: I think it'll be a better UX if we set the width of the `Box` to `fit-content` and add a background, so when the visitors play with the text they can see how the Box shrinks and grows, and also see the dimensions text change accordingly.

nikolovlazar (changes requested):
Thanks for opening this PR, @TylerAPfledderer! I've left you some notes about the contents. Let's fix them and we can merge.

pages/docs/styled-system/utility-hooks/use-dimensions.mdx
TylerAPfledderer: @nikolovlazar do you think it would also help to add a note above the example to encourage editing the text?

nikolovlazar: Don't think so. It says "EDITABLE EXAMPLE" on top of the code block.

TylerAPfledderer: Sounds good!

PR #428, opened 2022-03-14
docs: fix Getting Started > Remix `react` imports

imballinst:
Closes N/A (no existing issue)

## 📝 Description

This PR fixes the imports in the Remix Getting Started guide.

As far as I know, the `... as React` is used when we import all named exports using `*`, e.g.:

```ts
import * as React from 'react';
```

Instead, here, previously what we did was:

```ts
import React, { createContext } as React from "react"
```

This causes syntax errors in the IDE. In the later sections, we can see that the imports are specified differently than the ones above (which I think, this is the correct one).

```ts
// root.tsx
import React, { useContext, useEffect } from 'react';
```

## 🚀 New behavior

1. Remove the `as React` part of the line of code
2. Small update: use a single quote instead of double (to make it consistent with the later sections).

No

## 📝 Additional Information

If the previous one's syntax was valid, please let me know and I'll close this PR. Thanks!

### DIFF
--- a/pages/guides/getting-started/remix-guide.mdx
+++ b/pages/guides/getting-started/remix-guide.mdx
@@ -23,7 +23,7 @@ We’ll create a `context.tsx` in the `app` folder.
 
 ```jsx live=false
 // context.tsx
-import React, { createContext } as React from "react"
+import React, { createContext } from 'react'
 
 export interface ServerStyleContextData {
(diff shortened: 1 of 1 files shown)


nikolovlazar:
Good catch, @imballinst! Thank you for opening a PR.

PR #439, opened 2022-03-15
fix(get-search-meta): fix bug where local directory path is not being removed from url property in Windows 10

### LINKED ISSUE #438: Search Meta Generation Rendering Local Current Directory to JSON file on Windows 10 OS
### Description

When running `yarn search-meta:gen`, it will render a value in the `url` property that includes the local filepath to the pages.

i.e. `C:/Users/path/to/root/project/<page>`

### Proposed solution

The code in question is on line 37 in `scripts/get-search-mets.ts` with the `replace` method on `process.cwd()`. Because Windows OS can return backslashes instead of forward slashes, this command should be wrapped in the `fileToPath` function, which will swap the slashes.

So...
```js
  const slug = fileToPath(file)
    .replace(`/${websiteRoot}`, '')
    .replace(fileToPath(process.cwd()), '')
```

This should ensure consistency in matching the local directory path characters with what is being passed through the `file` parameter, no matter the OS.

TylerAPfledderer:
Closes #438 

## 📝 Description

On Windows 10, running `yarn search-meta:gen` script to generate an updated `search-meta.json` file will render the local directory path in the `url` property. This request wraps the `fileToPath` function around `process.cwd()` when trimming the slug to ensure the same slash characters in the filepath from the string passed through the `file` param of `getMDXMeta` and the current working directory string from node process.

Running `yarn search-meta:gen` renders in the `url` property the full local directory path in Windows 10...

i.e. `C:/Users/user/path/to/file/<page>`

where the meta file string passed to the `getMDXMeta` function contains forward slash characters, and `process.cwd()` returns a local directory path containing back slash characters.

## 🚀 New behavior

Passing `process.cwd()` to `fileToPath` when removing the local directory from the slug ensures that the slashes will match.

No

### DIFF
--- a/scripts/get-search-meta.ts
+++ b/scripts/get-search-meta.ts
@@ -35,8 +35,8 @@ async function getMDXMeta(file: string) {
   const tableOfContent = toc(content)
   const json = tableOfContent.json as TOCResultItem[]
   const slug = fileToPath(file)
-    .replace(`/${websiteRoot}`, '')
-    .replace(process.cwd(), '')
+    .replace(`${websiteRoot}`, '')
+    .replace(fileToPath(process.cwd()), '')
 
   const result: ResultType[] = []
 


nikolovlazar:
Good one, @TylerAPfledderer. I don't have a Windows machine to test this out. Maybe @aacevski can help review this one.

aacevski:
Looks good to me! We should try this on Mac tomorrow and we can merge it.

nikolovlazar:
I tried it on mac, no side-effects are happening. Good to go!

PR #452, opened 2022-03-19
fix(get-search-meta): Ensure incoming file string to getMDXMeta has forward slashes.

### LINKED ISSUE #449: Search -> FAQ -> Is there a Datepicker leads to 404
### Subject

Search

### Description

When searching "Is there a Datepicker in Chakra UI?" you will find an entry in the search but when clicking on the entry you will end on a 404.

https://chakra-ui.com/faq/index#is-there-a-datepicker-in-chakra-ui

Screen Shot 2022-03-18 at 3 49 56 PM
Screen Shot 2022-03-18 at 3 50 41 PM

I assume we either want to remove the entry or link it correctly if there is a document for this.

TylerAPfledderer:
## 📝 Description

Found issue related to contributors using Windows OS where incoming meta file strings into the `getMDXMeta` function can render backslashes instead of forward slashes. This prevents the `fileToPath` function from returning certain strings correctly.

Impacts search meta for the `Changelog` page as the rendered url strings included `index` at the end, resulting in 404 errors on search.

## 🚀 New behavior

Creating a variable to ensure all backslashes were forward slashes from the incoming file string prior to it being used allows the generated urls for the changelog page to not render `index` in the filepath.

No

## Additional Information

- Merged Pull Request #423 was impacted because the proper `search-meta.json` from that commit was overwritten.
- This should also close #449 as this fix addresses the 404 for that FAQ section.


TylerAPfledderer:
1. @nikolovlazar I must be an edge case because this was a problem previously; I am terribly sorry! 😅
2. @aacevski can you please verify I have made an appropriate and successful change since this is Windows related? 🙏‍

nikolovlazar:
LGTM! Let's see what @aacevski says after he tests this one.

aacevski:
Hey! This works well on my PC but I'm a bit confused on what the issue was in the first place? I've regenerated the search meta and for the MDX files the path stayed the same.

TylerAPfledderer:
@aacevski Anytime I tried to regenerate the meta, I was noticing with the changelog page url, say, it still added `index` at the end which creates a 404:

```js
  {
    "content": "Changelog",
    "id": "f1daa5ef-93c7-42bf-bb2d-64a172635e73",
    "type": "lvl1",
    "url": "/changelog/index",
    "hierarchy": { "lvl1": "Changelog" }
  },
```

When I was checking out the incoming file string from the `file` param of `getMDXMeta`, it was showing a string with backslashes, which interferes with the `fileToPath` function's ability to remove `index` on line 37.

_In the fileToPath function:_
```js
  if (indexRE.test(file)) {
    return file.replace(indexRE, '/$1');
  }
```
Where the `indexRE` regex is `/(^|.*\/)index\.(md|mdx|js|jsx|ts|tsx)$/i;`

TylerAPfledderer:
The backslashes are appearing in the `files` variable on line 73, as there is a `process.cwd()` method being called there.

I'm having trouble finding any specific discussion of this as it relates to Windows 10 and Node, other than generalized discussions about filepaths can be rendered with backslashes.

TylerAPfledderer:
I haven't found any discussions that are specific to my situation with Windows 10 and Node.

But I dove back into the `@docusaurus/utils` file and found the [`posixPath`](https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-utils/src/pathUtils.ts#L46-L68) function that handles just this case!

So I could do this instead:

```js
async function getMDXMeta(file: string) {
  
  const posixFile = posixPath(file) //<<<----- Here

  const { content, frontMatter: _frontMatter } = await parseMarkdownFile(file)
  const frontMatter = _frontMatter as Record<string, any>
  const tableOfContent = toc(content)
  const json = tableOfContent.json as TOCResultItem[]
  const slug = fileToPath(posixFile)
    .replace(`/${websiteRoot}`, '')
    .replace(posixPath(process.cwd()), '') // <<<------- And also here
```

Would this be acceptable, or too much?

nikolovlazar:
@TylerAPfledderer try it out. See if it helps in your case on Windows.

TylerAPfledderer:
@TylerAPfledderer it works for me! I think it's better to utilize a function from an existing package that serves this specific purpose, and can use it to keep the code tidy and consistent.

I've added a couple of other changes for consistency, and will push for review.

scripts/get-search-meta.ts
```diff
@@ -30,13 +35,19 @@ interface TOCResultItem {
 const websiteRoot = 'pages'
 
 async function getMDXMeta(file: string) {
-  const { content, frontMatter: _frontMatter } = await parseMarkdownFile(file)
+  // For Windows: convert backslashes to forwards slashes for consistency
+  const posixFile = posixPath(file)
+  const posixCWD = posixPath(process.cwd())
```
aacevski, line 40: I'd suggest naming these to:
- `pathToFile`
- `processCwd`
or something of those sorts, since I feel like these names would be hard to understand in the future.

aacevski:
Other than that, it works well for me on Windows!

scripts/get-search-meta.ts
TylerAPfledderer: That I will do! Thank you @aacevski.

TylerAPfledderer:
@nikolovlazar should be good to go at this point!

nikolovlazar:
@TylerAPfledderer I'm not getting the `useRadio` or `useRadioGroup` in the search meta. Can you try updating it again before we merge it?

TylerAPfledderer:
Got it, @nikolovlazar! Check again. 👍‍

nikolovlazar:
@TylerAPfledderer looks like it's ready to be merged! Just take care of the conflicts before we do that.

TylerAPfledderer:
@nikolovlazar conflicts resolved!

PR #455, opened 2022-03-20
Update number-index.mdx | TypeScript throws error

georgekrax:
TypeScript throws errors if we use `isReadOnly: true` in the `getInputProps()` function. Although, if we pass `isReadOnly` does work, there seems to be a type error. However, when using `getInputProps({ readOnly: true });` both TypeScript and functionalities work. So, there is not a change. You just have a stricter type system.

Maybe, change the initial types of the `getInputProps()` parameters?

Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior

### DIFF
--- a/pages/docs/components/form/number-input.mdx
+++ b/pages/docs/components/form/number-input.mdx
@@ -294,7 +294,6 @@ function HookUsage() {
       min: 1,
       max: 6,
       precision: 2,
-      isReadOnly: true,
     })
 
   const inc = getIncrementButtonProps()
--- a/pages/docs/components/form/number-input.mdx
+++ b/pages/docs/components/form/number-input.mdx
@@ -294,11 +294,12 @@ function HookUsage() {
       min: 1,
       max: 6,
       precision: 2,
+      isReadOnly: true,
     })
 
   const inc = getIncrementButtonProps()
   const dec = getDecrementButtonProps()
-  const input = getInputProps({ readOnly: true })
+  const input = getInputProps()
 
   return (
     <HStack maxW='320px'>
--- a/pages/docs/components/form/number-input.mdx
+++ b/pages/docs/components/form/number-input.mdx
@@ -298,7 +298,7 @@ function HookUsage() {
 
   const inc = getIncrementButtonProps()
   const dec = getDecrementButtonProps()
-  const input = getInputProps({ isReadOnly: true })
+  const input = getInputProps({ readOnly: true })
 
   return (
     <HStack maxW='320px'>


TylerAPfledderer:
Both options work, but this change is consistent with the [typing supplied to the function](https://github.com/chakra-ui/chakra-ui/blob/main/packages/number-input/src/use-number-input.ts#L509-L515), and `isReadOnly` is supposed to be supplied as an optional prop to `useNumberInput`.

georgekrax:
> Both options work, but this change is consistent with the [typing supplied to the function](https://github.com/chakra-ui/chakra-ui/blob/main/packages/number-input/src/use-number-input.ts#L509-L515), and `isReadOnly` is supposed to be supplied as an optional prop to `useNumberInput`.

Maybe update the type system then?

TylerAPfledderer:
> > Both options work, but this change is consistent with the [typing supplied to the function](https://github.com/chakra-ui/chakra-ui/blob/main/packages/number-input/src/use-number-input.ts#L509-L515), and `isReadOnly` is supposed to be supplied as an optional prop to `useNumberInput`.
> 
> Maybe update the type system then?

That question I have been wondering myself, and I believe is also related to issue [Chakra Issue #5572](https://github.com/chakra-ui/chakra-ui/issues/5722) with a similar function in the `useRadio` hook. Maybe the original intention was to not alter the `getInputProps` directly, and instead pass props to the hook itself which updates the returned functions.

But in this case, I think you are right in that the example should have `readOnly` and not `isReadyOnly` in the way it is being used. Else, else pass the prop into the hook instead.

```jsx
const { getInputProps, getIncrementButtonProps, getDecrementButtonProps } =
    useNumberInput({
      step: 0.10,
      defaultValue: 1.50,
      min: 1,
      max: 6,
      precision: 2,
      isReadOnly: true
    })

const inc = getIncrementButtonProps()
const dec = getDecrementButtonProps()
const input = getInputProps()
```

georgekrax:
@TylerAPfledderer So, should this be merged?

TylerAPfledderer:
That's not my call. 😅 The main member @nikolovlazar reviews most of these pull requests.

(I've been checking out the current PRs and wanted to include a note in agreement with your request! 😁)

nikolovlazar:
You're right, @TylerAPfledderer! We should provide the props to the hook instead of the prop getter functions. To fix the docsite, I recommend @georgekrax move the `readOnly` prop in the hook.

Also, adding the `readOnly` makes the input read-only, but you can still change its values by using the buttons. To make the input truly "read only" you need to pass the prop to the hook, so it will affect the buttons as well.

Let's move the `isReadOnly` prop to the hook instead.

nikolovlazar:
Hey @georgekrax! Thank you for the LinkedIn shoutout! As for this PR, do you need any help around it?

nikolovlazar:
@georgekrax I meant move the `isReadOnly` prop from the `getInputProps` to the `useNumberInput` hook here in the docs. I tested your PR in the main repo, but it still doesn't work. When a user sets the `isReadOnly` prop they expect the input to not change in any way, but if you click on the increment/decrement buttons you're still able to change the value. Users should not be setting the `isReadOnly` prop to the `getInputProps`, but to the `useNumberInput` hook instead to make their number input component truly read-only. I don't think we need to change the main repo, just change the guide here in the docs.

georgekrax:
@nikolovlazar Oh yeah, that is a better explanation. Thanks!

However, if we update the docs hook usage example to this:
```diff
function HookUsage() {
  const { getInputProps, getIncrementButtonProps, getDecrementButtonProps } =
    useNumberInput({
      step: 0.01,
      defaultValue: 1.53,
      min: 1,
      max: 6,
      precision: 2,
+     isReadOnly: true,
    })

  const inc = getIncrementButtonProps()
  const dec = getDecrementButtonProps()
- const input = getInputProps({ readOnly: true })

  return (
    <HStack maxW='320px'>
      <Button {...inc}>+</Button>
      <Input {...input} />
      <Button {...dec}>-</Button>
    </HStack>
  )
}
``` 

Then, the user will not be able to update the number input value, neither through the increment/decrement buttons nor through the input. So, I guess the best is to leave it as it is?

TylerAPfledderer:
@georgekrax you still have to have `const input = getInputProps()` to spread into the `Input` component. You just need to remove `{ readOnly: true}` from it.

georgekrax:
@TylerAPfledderer In my code I do not use `{ readOnly: true }`. However, there might be a reason why @segunadebayo left it there. He created this example for a **Mobile Spinner**, so I guess he did not want the user to change the input value directly in mobile devices

nikolovlazar:
@TylerAPfledderer it is. Setting the `isReadOnly` prop in the `useNumberInput` will make not just the input, but the whole component read-only, so buttons will be affected as well. I know it's not explicitly specified that it affects the button as well, but the intention of the prop is exactly that.

georgekrax:
I suggest we close the Pull request, as it does not help and make things more complicated.

nikolovlazar:
@georgekrax Just delete the `{ readOnly: true }` from the `getInputProps` and add the `isReadOnly` to the `useNumberInput` hook.

georgekrax:
@nikolovlazar I understand that, but why the heck would you want the whole component to be readonly. Just add the `isReadOnly` to the `useNumberInput()` hook in your docs demo code preview, and you will see that the user cannot update at all the number value, neither through the buttons nor through the actual input.

nikolovlazar:
@georgekrax that's the point. If you add `isReadOnly` it should truly be read-only, the whole component not just the input. Having a use case where only the input is read-only but the buttons aren't affected is an edge case, so the modifications should be on the user-land, not built in Chakra UI. Don't think about that edge case. When we add `isReadOnly` we expect the input along with the buttons to be read only.

georgekrax:
@nikolovlazar And you want the whole component to be "disabled"? That is weird.

nikolovlazar:
> @nikolovlazar And you want the whole component to be "disabled"? That is weird. Why would you want the component to be disabled in the docs, should not the user and the developer be able to interact with it?

That's the point of the isReadOnly prop. Also, it's a live editable preview, so anyone can remove the isReadOnly prop and interact with the component.

georgekrax:
@nikolovlazar Okay, whatever you like. I just do not want someone else to open another **Pull request**, stating that they have to remove the `isReadOnly: true` parameter in order for the component to be functional.

nikolovlazar:
I think that part needs rework. Here's what we can do, remove the read only props totally for now, but we can write a better documentation for the hook. How does that sound?

georgekrax:
> I think that part needs rework. Here's what we can do, remove the read only props totally for now, but we can write a better documentation for the hook. How does that sound?

Yes, that is what I am talking about.

PR #462, opened 2022-03-22
docs: delete unnecessary import in remix template

aiji42:
## 📝 Description

There is a description to import dotenv in the entry.server.tsx template, but this is not necessary.

スクリーンショット 2022-03-22 午後4 25 42

- This does not affect the rendering of chakra-ui.
- For chakra-ui beginners, it may be a useless guess that some implicit environment variable definitions are necessary.
    - Yes, I had an evil guess.
- If the runtime is cloudflare, it cannot be built because it is not Node and process cannot be referenced.


nikolovlazar:
Makes total sense, @aiji42! Thank you for opening this PR.

@allcontributors please add @aiji42 for docs.

PR #471, opened 2022-03-24
docs/adding blog 

estheragbaje:
## 📝 Description

This PR adds in the blog layout and  a blog post

At the moment, there's no blog on the site

## 🚀 New behavior

Introduced a blog layout and added a blog post

No


nikolovlazar:
Love it! Let's bring the blog back! 🚀

Also, I noticed you're missing the search meta and sidebar link.

estheragbaje:
@nikolovlazar Yup! We're bringing back the blog and hoping it stays active.

My thought is that we should have a separate search for blogs different from the general doc search.  What do you say?

Wanted to push this first post out and then add in the sidebar links subsequently.

nikolovlazar:
That would make sense as well. I'll think more about this later today.

noobinthisgame:
@estheragbaje Nice work! 

Two things that went through my mind: 
- Are only people from the chakra team allowed to add blog entries ? 
- We should really start to think of a better way to structure the sidebar and the main nav links. If you plan to add the blog on main nav links we should remove something else instead. Thinking forward to the tutorial section that should have a good place as well, but will make it more crowded.

nikolovlazar:
Good points, @noobinthisgame. Also, I guess we also need to build a mini "landing page" for the blog, where we can put all of our articles in a grid/gallery.

estheragbaje:
Thanks @noobinthisgame! The goal is to have other contributors to the blog asides from the Chakra team though I  structured it that way cos the author's details were already available... Can rethink this and see how to make it more flexible.

I agree, intend to create a PR later to add the mini landing page for the blog

noobinthisgame:
@estheragbaje we can go with this for now and when someone new wants to add something we can think of another way for details. Or maybe build a fallback for when there is no informations ?

estheragbaje:
@noobinthisgame That's what I think too. Let's go with this for now 🚀

nikolovlazar:
Sounds good, @estheragbaje! I'll merge this just so we can have a link of the article to share. And I like the game plan:
- Open another PR for the mini landing page
- Add the blog mini landing page in the sidebar
- Figure out how we can have a separate search for the articles

PR #488, opened 2022-03-29
docs: wrap examples with overflow issues in the `TableContainer` component

TylerAPfledderer:
## 📝 Description
This pull request surveys live code previews site-wide and wraps the code in the `TableContainer` component so the previews do not overflow the container when viewing in mobile screens.

This does not address any issues with the examples themselves not rendering as intended in mobile screens.

Found sections to fix:
- [X] [Flex vs Grid vs HStack](https://chakra-ui.com/docs/components/layout/flex#flex-and-spacer-vs-grid-vs-stack)
- [x] [Radio Group Usage](https://chakra-ui.com/docs/components/form/form-control#sample-usage-for-a-radio-or-checkbox-group)
- [x] [Changing Number Input Size](https://chakra-ui.com/docs/components/form/number-input#changing-the-size-of-the-input)
- [x] [Code Colors](https://chakra-ui.com/docs/components/data-display/code#colors)
- [x] [Custom Toast Container](https://chakra-ui.com/docs/components/feedback/toast#custom-container-styles)
- [x] [Tooltip Placement](https://chakra-ui.com/docs/components/overlay/tooltip#placement)
- [x] [Image Usage](https://chakra-ui.com/docs/components/media-and-icons/image#usage) and [Size](https://chakra-ui.com/docs/components/media-and-icons/image#size)

One or more live preview examples in the docs overflow their `react-live` preview container when viewed on a mobile screen

## 🚀 New behavior

The wrapping `TableContainer` component with render an `overflow-x: auto` to the offending examples to create a horizontal scroll.

No

## 📝 Additional Information

Still in draft as all doc pages have not been reviewed. Description of the pull request may change based on further findings.

Using `TableContainer` instead applying a global `overflow: auto` style to the react-live preview component was recommended by the docsite maintainer.

Consider this PR to be an interim temporary solution to at least some of the examples in question that need to be redone. (In that case, open separate issues/PRs for each example)

### DIFF
--- a/pages/docs/components/data-display/code.mdx
+++ b/pages/docs/components/data-display/code.mdx
@@ -8,8 +8,8 @@ description:
 ---
 
 `Code` is a component used to display inline code. It is composed from the
-[Box](/docs/components/layout/box) component with a font family of `mono` for displaying
-code.
+[Box](/docs/components/layout/box) component with a font family of `mono` for
+displaying code.
 
 <ComponentLinks
   theme={{ componentName: 'code' }}
@@ -37,11 +37,14 @@ You can change the color scheme of the component by passing the `colorScheme`
 prop.
 
 ```jsx
-<Stack direction='row'>
-  <Code children='console.log(welcome)' />
-  <Code colorScheme='red' children="var chakra = 'awesome!'" />
-  <Code colorScheme='yellow' children='npm install chakra' />
-</Stack>
+// TableContainer is used as a wrapper to prevent overflow in mobile screens
+<TableContainer>
+  <Stack direction='row'>
+    <Code children='console.log(welcome)' />
+    <Code colorScheme='red' children="var chakra = 'awesome!'" />
+    <Code colorScheme='yellow' children='npm install chakra' />
+  </Stack>
+</TableContainer>
 ```
 
 ## Props
--- a/pages/docs/components/feedback/toast.mdx
+++ b/pages/docs/components/feedback/toast.mdx
@@ -240,17 +240,20 @@ function Example() {
   // Or via trigger
   // Style here will overwrite the entire style above
   return (
-    <Button
-      onClick={() => {
-        toast({
-          containerStyle: {
-            border: '20px solid red',
-          },
-        })
-      }}
-    >
-      Click me to show toast with custom container style.
-    </Button>
+    // TableContainer is used as a wrapper to prevent overflow in mobile screens
+    <TableContainer>
+      <Button
+        onClick={() => {
+          toast({
+            containerStyle: {
+              border: '20px solid red',
+            },
+          })
+        }}
+      >
+        Click me to show toast with custom container style.
+      </Button>
+    </TableContainer>
   )
 }
 ```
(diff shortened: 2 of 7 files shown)


noobinthisgame (commented):
This looks good to me! I would only put the comment for `Tablecontainer` in my one place. But that's personal preference I guess.

TylerAPfledderer:
> This looks good to me! I would only put the comment for `Tablecontainer` in my one place. But that's personal preference I guess.

@noobinthisgame yea I thought about that.

But if anyone added a new example that had the potential of overflow, I would think to also add a PR to the Contributing file and page to add instructions to add the `TableContainer` component and the visual comment.

nikolovlazar:
I thought about this and I think we need to be smarter about it. I know it fixes the issue, but here are some cons that I have thought of:
- All examples are now "tainted" with the `TableContainer`, plus the comment explaining why it's there. With that we're also damaging the "copy/paste" experience, because the users will need to clean it up after pasting the example in their project.
- Some beginners could get confused because they're seeing a component named `TableContainer` in the Tooltip docs for example.
- If we have a problem with the responsiveness, the users should not be affected by the solution. This solution is more of a compromise.

Why don't we just modify the CodeBlock component in `src/components/mdx-components/codeblock/codeblock.tsx`? That way all of the examples will have the responsiveness fix that the table container offers, but it won't be part of the examples and won't be visible to the users.

TylerAPfledderer:
@nikolovlazar agreed! Probably better to close this PR and open a fresh one?

TylerAPfledderer:
Also, @nikolovlazar, just to clarify, I was thinking adding overflow to this component would also work?

https://github.com/chakra-ui/chakra-ui-docs/blob/main/src/components/mdx-components/codeblock/react-live-block.tsx#L10

TylerAPfledderer:
All good, @noobinthisgame! It was still good to show the `TableContainer` in those examples for best practice. 😄

nikolovlazar:
Yeah I think so too @TylerAPfledderer. The example you shared does have a longer text and on mobile it goes out of the preview box. The `ReactLiveBlock` component is actually a wrapper and it exists in the project here:

`src/components/mdx-components/codeblock/react-live-block.tsx`

The preview is rendered in the `<LiveCodePreview zIndex='1' />`. You can wrap it with a box and apply the style props you think it'll solve the overflow issue.

TylerAPfledderer:
@nikolovlazar so wrapping `<LiveCodePreview zIndex='1' />` with a box like this:
```jsx
<Box overflowX='auto' whiteSpace='pre'>
  <LiveCodePreview zIndex='1' />
</Box>
``` 

Shows this:
image

I believe applying the changes to line 10 in `react-live-block.tsx` will solve this issue since they will be applied within `LiveCodePreview`
```jsx
const LiveCodePreview = chakra(LivePreview, {
  baseStyle: {
    fontFamily: 'body',
    mt: 5,
    p: 3,
    borderWidth: 1,
    borderRadius: '12px',

    overflowX: 'auto',
    whiteSpace: 'pre'
  },
})
```
image

nikolovlazar:
@TylerAPfledderer that's the solution! Apply those changes, clean up the PR and let me know when it's ready to be reviewed.

TylerAPfledderer:
Closing this pull request to avoid polluting the commit history with bad commits. Opening a new PR to push the proper changes for the same overflow issue.

Issue #192, opened 2022-01-01
docs(framer-motion): Update "Chakra UI + Framer Motion" integration guide's examples to use the "chakra" factory + "as" prop


tannguyencse19:
### Subject

Code example

### Description

Why don't use like this?

```jsx
<Box as={motion.div} />
```

**_Cons_**: `chakra` props will be prioritized over `framer-motion` props if existing prop that have the same name. e.g:

```jsx
<Box 
    as={motion.div} 
    transition="all 0.5s linear" 
   /* not: 
   transition={
        duration: 0.5,
        ease: "linear"
   }
   */
/>
```

noobinthisgame:
Hi @tannguyencse19, could you elaborate on this? To continue further here we need more details on what you are missing and what/where things have to change in the docs.

nikolovlazar:
@noobinthisgame I think @tannguyencse19 has a point. The docs utilize the `motion` factory function, which has changed over the time and it's throwing an error now. We could utilize the `chakra` factory function, just like we do when wrapping the Next.js Image component. Here's a codesandbox that implements the two ways: https://codesandbox.io/s/eager-estrela-wwg3k4?file=/src/index.tsx. The behaviour is the same, but there's an error in the `motion` way. I guess we should update the docs to use the `chakra` factory function instead.

wattanx:
I'd like to work on this!

tannguyencse19:
> @noobinthisgame I think @tannguyencse19 has a point. The docs utilize the `motion` factory function, which has changed over the time and it's throwing an error now. We could utilize the `chakra` factory function, just like we do when wrapping the Next.js Image component. Here's a codesandbox that implements Framer Motion in three different ways: https://codesandbox.io/s/eager-estrela-wwg3k4?file=/src/index.tsx. The behaviour is the same, but there's an error in the `motion` factory function way. I guess we should update the docs to use the `chakra` factory function and the `as` prop instead:
> 
> ```ts
> // `chakra` factory function
> const ChakraBox = chakra(motion.div, {
>   shouldForwardProp: isValidMotionProp
> });
> ```
> 
> or
> 
> ```js
> // `as` prop
> <Box
>    as={motion.div}
>    height="40px"
>    width="40px"
>    bg="orange.400"
>    drag="x"
>    dragConstraints={{ left: -100, right: 100 }}
>    whileHover={{ scale: 1.1 }}
>    whileTap={{ scale: 0.9 }}
> />
> ```

[...]

PR #493, opened 2022-04-01
Added other methods to the Chakra UI + Framer Motion section.

wattanx:
Closes https://github.com/chakra-ui/chakra-ui-docs/issues/192

## 📝 Description
Added other methods to the `Chakra UI + Framer Motion` section.
use the `chakra factory` + `as props`

Only `motion factory`

## 🚀 New behavior

added `chakra factory` + `as props`

no


wattanx:
@nikolovlazar 
There is one thing that concerns me.
When I tried, `motion factory` did not result in an error.
(It looks like `@types/react` and others are probably affected.)
https://codesandbox.io/s/pedantic-thunder-q49e9y?file=/src/index.tsx

I am wondering if I can keep the `motion factory` as well, what do you think?

nikolovlazar:
Hi @wattanx. I just tried creating a new typescript CRA, copy pasted the motion example and I do get the error on the `.custom` method (`motion.custon is not a function`). The Chakra Factory Function method does the same without any errors, so let's get rid of the motion factory function, just keep the Chakra one.

wattanx:
@nikolovlazar
Thank you!
I removed `motion factory` because the same thing can be done with chakra factory.

Issue #401, opened 2022-03-07
Outdated/confusing docs for setting Container maxWidth in theme.sizes.container


webwelten:
### Subject

component

### Description

I must admit that I'm a little confused what's the best way/best practice to use the Container. 
In the Container docs it says:

> By default, it sets the max-width of the content to 60 characters (60ch) but you can customize this by passing custom maxWidth values **or changing the container size tokens defined in theme.sizes.container**

First of all: I've never been in a project that used the character size as maxWidth (although I can see how it makes sense). Often frameworks would offer maxWidth values for certain breakpoints instead (like bootstrap etc. for example) or in projects we would follow the designers guidelines. Do you have a different experience? What's the reason behind this default? (I am just curious. As long as I can adapt it, I'm fine with this being the default)

> you can customize this by passing custom maxWidth values **or changing the container size tokens defined in theme.sizes.container**

This sounds to me that changing **theme.sizes.container** would adapt every Container I use. But instead it just changes what I can reference in maxW, right? I find that part a little confusing and I think that adapting maxW everytime is more an exception. I would expect that the docs provide a best practice for the global option and only then describes how to adapt certain cases.

As far as I know my use case can be solved like that:

```
components: {
  Container: {
    baseStyle: {
      maxWidth: […]
    }
  }
}
```

Did you get my point? Any other perspectives or other ways your using Container?

noobinthisgame:
Hey @webwelten ! 
I don't know what exactly was the reasoning behind the default value (maybe @segunadebayo can give you more insights about that). In any case the maxWidth in the `Container` is adjustable for your needs and use cases.

Overriding the values of the `theme.sizes.container` in your custom theme, changes the value of the tokens used in the maxWidth props as in this [example](https://chakra-ui.com/docs/components/layout/container#container-size). This would be my preferred approach You can find the default values in this file: https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/foundations/sizes.ts#L23 

Your solution of changing the theme for the `Container` through a custom theme is also a valid way to address those changes. 

Feel free to raise a PR with improvement to the docs. 😊🚀

TylerAPfledderer:
I can comment that the default value of `60ch` is the common approach to the best readability of a single column of text on desktop (a range of 45-75 characters per line, to be more precise).

You have to keep in mind though that this unit is based on the width of the character `0` which depends on the typeface used. Maybe it depends on context if this creates undesirable results when using more than one column?

[...]

PR #497, opened 2022-04-03
docs(container): update "Container Size" section

TylerAPfledderer:
Closes Issue #401

## 📝 Description

Moves the details about `maxWidth` to the section [Container Size](http://localhost:3000/docs/components/layout/container#container-size) and update the description and examples

- The detail about the default `maxWidth` sits at the top of the page away from the size section
- The container size live code example has no color to be able to easily visualize the different maxWidth values rendered to the `Container` components.
- Per discussion in issue #401, there is no explanation on the page given about the default value of `60ch`. Not everyone knows about this unit value or knows why this particular value might be used. (This is related to UX/UI best/suggested practice)

## 🚀 New behavior

- The details about the default `maxWidth` are moved to the sizes section and updated to include notes about the default value, providing a definition about the unit value used and a source to view to consider the purpose of the given default value of `60` characters.
- The live code example is simplified to show an example of each type a value that can be used:
  - a `size` token
  - a `size.container` token
  - an explicit unit value
- Change the text `theme.sizes.container` in the details to `theme.sizes`, and convert it to a link to the style props section in the docs.

No

### DIFF
--- a/pages/docs/components/layout/container.mdx
+++ b/pages/docs/components/layout/container.mdx
@@ -61,7 +61,7 @@ values or changing the size tokens defined in
     "550px" Container
   </Container>
   <Container maxW='container.sm' bg='green.400' color='#262626'>
-    "small" Container
+    "container.sm" Container
   </Container>
 </VStack>
 ```
--- a/pages/docs/components/layout/container.mdx
+++ b/pages/docs/components/layout/container.mdx
@@ -46,10 +46,15 @@ tokens defined in
 
 ```jsx
 <VStack>
-  <Container maxW='container.xl'>Extra-Large Container</Container>
-  <Container maxW='container.lg'>Large Container</Container>
-  <Container maxW='container.md'>Medium Container</Container>
-  <Container maxW='container.sm'>Small Container</Container>
+  <Container maxW='md' bg='blue.600' color='white'>
+    "md" Container
+  </Container>
+  <Container maxW='550px' bg='purple.600' color='white'>
+    "550px" Container
+  </Container>
+  <Container maxW='container.sm' bg='green.400' color='#262626'>
+    "small" Container
+  </Container>
 </VStack>
 ```
 
--- a/pages/docs/components/layout/container.mdx
+++ b/pages/docs/components/layout/container.mdx
@@ -39,9 +39,9 @@ To contain any piece of content, wrap it in the `Container` component.
 
 ## Container Size
 
-By default, it sets the `max-width` of the content to 60 characters (`60ch`) but
-you can customize this by passing custom `maxWidth` values or changing the size
-tokens defined in
+By default, the `Container` component sets the `maxWidth` of the content to 60
+characters (`60ch`) but you can customize this by passing custom `maxWidth`
+values or changing the size tokens defined in
 [theme.sizes](https://chakra-ui.com/docs/styled-system/theming/theme#sizes).
 
 ```jsx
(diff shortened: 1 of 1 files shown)


aacevski:
LGTM, great work @TylerAPfledderer 🚀

pages/docs/components/layout/container.mdx
```diff
@@ -43,14 +39,30 @@ To contain any piece of content, wrap it in the `Container` component.
 
 ## Container Size
 
-To set the size of a Container, use the `maxW` attribute.
+By default, the `Container` component sets the `maxWidth` of the content to 60
+characters (`60ch`) but you can customize this by passing custom `maxWidth`
+values or changing the size tokens defined in
+[theme.sizes](https://chakra-ui.com/docs/styled-system/theming/theme#sizes).
```
noobinthisgame, line 45: I'd leave this as before with theme.size.container to leave no room for confusion.

pages/docs/components/layout/container.mdx
```diff
@@ -43,14 +39,30 @@ To contain any piece of content, wrap it in the `Container` component.
...
-  <Container maxW='container.sm'>Small Container</Container>
+  <Container maxW='md' bg='blue.600' color='white'>
+    "md" Container
+  </Container>
+  <Container maxW='550px' bg='purple.600' color='white'>
+    "550px" Container
+  </Container>
+  <Container maxW='container.sm' bg='green.400' color='#262626'>
+    "container.sm" Container
+  </Container>
```
noobinthisgame, line 65: I'd also leave the old example as it was, as it was meant to show the different default sizes that we have for `container`. I'd add an example with custom `maxW` after mentioning that it's a way to style it.
The first paragraph could be about `maxW` then example and the second with sizes from theme with the old example.

noobinthisgame (commented):
Thanks for taking on this. I've left some thoughts :)

pages/docs/components/layout/container.mdx
TylerAPfledderer: Understandable. Easy enough! 😄

pages/docs/components/layout/container.mdx
TylerAPfledderer: Along with adding in a background color to each `Container` to be able to better see it's width, I was thinking that the original example didn't truly contribute anything, as some of the sizes from the `container` token are greater than the preview window's widest point (around a `767px` screen size), hence linking to the style props page via the `theme.sizes.container` text in the details or listing the container token sizes instead, and simplifying the example.

Thoughts?

noobinthisgame: Convinced :D Is the container.md working in the example ? At least to add a second one ?

TylerAPfledderer: `container.md` is wider than the other new examples with a max-width of `768`, so it will always show as maxed-out in the preview (I know, just shy! 🤣 The layout shift of the page happens at the 768 breakpoint). But I suppose that would be fine as an additional example. 😃

nikolovlazar: I don't think it's possible to showcase the container widths within the docsite itself. One thing we can do it setup a CodeSandbox example with the different widths and share the full page example (not the csb IDE link). That way the examples will be in full width of the screen and the users will be able to see the differences.

noobinthisgame: @nikolovlazar in that case I thin I'd just go with what @TylerAPfledderer did for now. And if we see people struggle we can add the csb example?

nikolovlazar: sounds good!

noobinthisgame (approved):

PR #498, opened 2022-04-03
docs(stack): Update Stack documentation

### LINKED ISSUE #465: (docs): add clarification to the `Stack` doc details/examples
### Subject

Stack

### Description

There are a couple of sections in the [`Stack` Component Docs](https://chakra-ui.com/docs/components/layout/stack) showing arguably redundant, incorrect, and/or confusing information.

- Section ["Stack Items Horizontally"](https://chakra-ui.com/docs/components/layout/stack#stack-items-horizontally): 
  - The details talk about using the `direction` prop, but the example uses the `HStack` component. This might also be a redundant section as it is currently being presented. 
  - Possible consideration would be to remove those contradicting details, and redoing the structure of the page so that this example is shown as part of a group with [the previous example](https://chakra-ui.com/docs/components/layout/stack) under a new header like `Use Case Examples`
```
### Use Case Examples
  #### Feature Cards with Stack Component
  #### Feature Cards with HStack Component
```

- Section ["Notes on Stack vs Flex"](https://chakra-ui.com/docs/components/layout/stack): 
  - The details here are not accurate and seem to be unfinished in providing a quality comparison. Combining `Stack` and `Flex` (I guess this means `<Stack as={Flex}>` or vice versa) does not allow you to alter spacing of any one child component -- via `margin`, say -- as that change will still be overwritten by the styling from `Stack`. You can use the `Spacer` component though to add spacing to the value provided by `Stack`.
  -  `justifyContent` is a valid prop for both components so it's not clear as to the purpose in including it in the explanation. 
  - Possible considerations are to remove this detail, and either referencing the `Flex` section ["Flex and Spacer vs Grid vs Stack"](https://chakra-ui.com/docs/components/layout/flex#flex-and-spacer-vs-grid-vs-stack), creating a more clear note in the `Stack` doc, or provide more clarity in both sections in the `Stack` and `Flex` docs to better articulate the differences.

Some other areas of this page may or may not provide enough or clear information, but the points above stuck out the most. I appreciate any thoughts to my observations. 😄

aacevski:
Closes #465 

## 📝 Description
This PR improves two sections of the Stack documentation. 
- The examples for horizontally stacking items.
- The section Flex vs Stack

Currently the above mention sections are contradictory and unclear.

## 🚀 New behavior

I've made sure to clear things up by providing more examples and updating the section regarding the difference between Flex & Stack.

No.

### DIFF
--- a/pages/docs/components/layout/stack.mdx
+++ b/pages/docs/components/layout/stack.mdx
@@ -138,6 +138,10 @@ render(<StackEx />)
 
 #### Feature Cards with HStack Component
 
+In the example following, we can notice that for the `HStack` component, we
+don't need the direction props, since this component is specifically for
+horizontally stacking items.
+
 ```jsx manual=true
 function Feature({ title, desc, ...rest }) {
   return (
--- a/pages/docs/components/layout/stack.mdx
+++ b/pages/docs/components/layout/stack.mdx
@@ -168,10 +168,9 @@ render(<StackEx />)
 
 ### Notes on Stack vs Flex
 
-Stack's primary use case is to lay items out and control the spacing between
-each item. If you have a more complicated use case, such as changing the margin
-on the last child, you can combine `Stack` and `Flex` and use
-`justify-content: space-between` for more control of the layout.
+The `Stack` component and the `Flex` component have their children spaced out
+evenly but the **key** difference is that the `Stack` won't span the entire
+width of the container whereas the `Flex` will.
 
 ## Props
 
--- a/pages/docs/components/layout/stack.mdx
+++ b/pages/docs/components/layout/stack.mdx
@@ -101,7 +101,12 @@ element.
 </VStack>
 ```
 
-### Example
+### Stack items horizontally
+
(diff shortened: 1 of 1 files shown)


pages/docs/components/layout/stack.mdx
```diff
@@ -131,22 +136,16 @@ function StackEx() {
...
 ```
 
-### Stack items horizontally
+#### Feature Cards with HStack Component
 
-Pass the `direction` and set it to `row`. Optionally, you can use `align` and
-`justify` to adjust the alignment and distribution of the items.
+In the example following, we can notice that for the `HStack` component, we
+don't need the direction props, since this component is specifically for
+horizontally stacking items.
```
noobinthisgame, line 143: You could also mention that the Items are also aligned in the center by default :)

noobinthisgame (changes requested):
Nice clarification! Check my comment and then it's good to go :)

aacevski:
Thanks a lot @noobinthisgame, I've added it as well!

noobinthisgame (approved):

PR #535, opened 2022-04-15
fix(alert): update second composition code example

TylerAPfledderer:
## 📝 Description

The second live code example for the `Alert` doc's [Composition section](http://localhost:3000/docs/components/feedback/alert#composition) shows a `CloseButton` component, but no example of functionality to assure the user that the removal of the alert happens with no props passed to the wrapper nor within the component, but through some other means outside of the scope.

This PR also tidies up the layout of the example in respect to mobile view, where the `CloseButton` component stays within the flow of the flexbox rendered by `Alert` and positioned relatively. This keeps any content from potentially overlapping this component should a user play around with the example.

The user might try to press the close button, but the alert will not close

## 🚀 New behavior

Include in the example code `useDisclosure` and handling of `CloseButton` on click so the the alert will disappear when this button is selected.


pages/docs/components/feedback/alert.mdx
```diff
@@ -132,20 +132,35 @@ kinds of layouts. Here's an example of a custom alert style and layout:
...
+          <AlertDescription>
+            Your application has been receivd. We will review your application
+            and respond within the next 48 hours.
+          </AlertDescription>
+        </Box>
+        <CloseButton
+          alignSelf='flex-start'
+          position='relative'
+          right='-4px'
+          top='-4px'
```
nikolovlazar, line 157: Let's use `{-1}` instead of `-4px`. It compiles to `-4px` as well, but it's much cleaner.

pages/docs/components/feedback/alert.mdx
```diff
@@ -132,20 +132,35 @@ kinds of layouts. Here's an example of a custom alert style and layout:
...
-    <AlertTitle>Success!</AlertTitle>
-    <AlertDescription display='block'>
-      Your application has been received. We will review your application and
-      respond within the next 48 hours.
-    </AlertDescription>
-  </Box>
-  <CloseButton position='absolute' right='8px' top='8px' />
-</Alert>
+function CompExample() {
+  const { isOpen, onClose } = useDisclosure({ defaultIsOpen: true })
```
nikolovlazar, line 140: I love that you expanded on this example and made it functional. How about adding a simple Toggle button that can bring up the alert again after we've closed it?

nikolovlazar (changes requested):

TylerAPfledderer:
@nikolovlazar good to go for a re-review! 🎉 🌮

PR #549, opened 2022-04-22
fix(with-framer): update factory example and details

### LINKED ISSUE #541: Forward children when using chakra factory with framer-motion
### Subject

Chakra UI + Framer Motion

### Description

It seems that `isValidMotionProp` does not take `children` into account. Since a motionized `Box` is used in the example, it would be  fitting to also add the `children` in the  `shouldForwardProp` function or add a notice for this.

TylerAPfledderer:
Closes #541

## 📝 Description

This is an update to the Chakra Factory example for [Chakra + Framer](https://chakra-ui.com/guides/integrations/with-framer#usage), where it shows the inclusion of the `children` prop in `shouldForwardProps`

Currently, `shouldForwardProp` only passes in `isValidMotionProp` so that the Framer animations are active and not overridden by matching props from Chakra. However, you can't use child components because the `children` props ends up being ignored, despite all other chakra props still available for use. This is not notated for users to be aware of.

## 🚀 New behavior

The inclusion of forwarding the `children` prop in `shouldForwardProps` is provided in the example with a comment underneath. All other chakra props not matching the motion props can still be used.

The overall example is updated to reflect the `children` prop being included.

### DIFF
--- a/configs/sandpack-contents/framer-motion-examples/chakra-factory.js
+++ b/configs/sandpack-contents/framer-motion-examples/chakra-factory.js
@@ -3,6 +3,10 @@ module.exports = {
   import { motion, isValidMotionProp } from 'framer-motion';
     
   const ChakraBox = chakra(motion.div, {
+    /**
+     * Allow motion props and the children prop to be forwarded.
+     * All other chakra props not matching the motion props will still be forwarded.
+     */
     shouldForwardProp: (prop) => isValidMotionProp(prop) || prop === 'children',
   });
     
--- a/configs/sandpack-contents/framer-motion-examples/chakra-factory.js
+++ b/configs/sandpack-contents/framer-motion-examples/chakra-factory.js
@@ -1,47 +1,51 @@
 module.exports = {
   App: `import { Container, chakra } from '@chakra-ui/react';
-import { motion, isValidMotionProp } from 'framer-motion';
-  
-const ChakraBox = chakra(motion.div, {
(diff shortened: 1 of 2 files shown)


configs/sandpack-contents/framer-motion-examples/chakra-factory.js
```diff
@@ -1,47 +1,51 @@
...
-  </ChakraProvider>,
-  rootElement
-);`,
-}
+module.exports = {
+  App: `import { Container, chakra } from '@chakra-ui/react';
+  import { motion, isValidMotionProp } from 'framer-motion';
+    
+  const ChakraBox = chakra(motion.div, {
+    shouldForwardProp: (prop) => isValidMotionProp(prop) || prop === 'children',
```
noobinthisgame, line 6: I would add a comment above here to make this bit point out more. Would make it easier to find after reading the hint bellow

noobinthisgame (changes requested):
This is looking good already. I've left one little thought.

configs/sandpack-contents/framer-motion-examples/chakra-factory.js
TylerAPfledderer: Done!

TylerAPfledderer:
@noobinthisgame PR has been updated! 😸

noobinthisgame:
Great !

noobinthisgame (approved):
Let's go !

PR #551, opened 2022-04-22
docs(flex): update spacer live example for mobile view

TylerAPfledderer:
## 📝 Description

Updates the section regarding the `Spacer` component and update the live code example to have a better view on a mobile device.

The layout of the example code at mobile view (smallest view of `375px`) looks broken, with a compressed "logo" text and tightly stacked buttons.

## 🚀 New behavior

Added documentation mentions the behavior of the `Spacer` with a parent using the `gap` prop.

The example code is refactored so that the buttons are appropriately wrapped in a `ButtonGroup` component, and the entire example does not collapse, with the intention that this example is meant for desktop view and not mobile view. Added a note about the latter.

### DIFF
--- a/pages/docs/components/layout/flex.mdx
+++ b/pages/docs/components/layout/flex.mdx
@@ -118,18 +118,23 @@ different widths differently.
 A good use case for `Spacer` is to create a navbar with a signup/login button
 aligned to the right.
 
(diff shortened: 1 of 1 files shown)


noobinthisgame (approved):
LGTM!

PR #565, opened 2022-04-28
docs: add yarn installation command for remix

Jannis-Morgenstern:
## 📝 Description

> Add installation command for remix using yarn. Could have changed the label "Inside your Remix project root directory, install Chakra UI by running the following:" to singular but chose to add instructions for yarn instead.

## 💣 Is this a breaking change (Yes/No): No


noobinthisgame (approved):
LGTM :)

PR #568, opened 2022-05-03
docs(use-disclosure): create details for the getter methods

### LINKED ISSUE #563: [useDisclosure hook] getButtonProps and getDisclosureProps are not documented
### Subject

useDisclosure

### Description

The `getButtonProps` and `getDisclosureProps` functions returned by the `useDisclosure` hook are not mentioned in the documentation, neither are they used in any code example. This is a pity, since they are quite handy :rocket:

Can you add some documentation on them with examples on how they could be used?

Thanks!

TylerAPfledderer:
Closes #563 

## 📝 Description

Add documentation for the getter methods returned in the `useDisclosure` hook, and update the rest of the docs for better accuracy of overall usage of the hook.

Documentation only provides examples of the functions and boolean value returned by the hook

## 🚀 New behavior

Adds the getter methods with their own example, showing their usage as the most basic with the hook. Updates the original example by providing detail on how the other methods and boolean value returned from the hook are used to provide more control.

## 📝 Additional Information

Until the docsite dependencies are bumped, the `getButtonProps` will still return a static `aria-expanded` value if the user inspects the example button via devtools. Fixed in PR [#5940](https://github.com/chakra-ui/chakra-ui/pull/5940).

### DIFF
--- a/pages/docs/styled-system/utility-hooks/use-disclosure.mdx
+++ b/pages/docs/styled-system/utility-hooks/use-disclosure.mdx
@@ -46,11 +46,11 @@ The component that uses `getDisclosureProps` receives the following props:
 
 The button that uses `getButtonProps` for toggling receives the following props:
 
-- An optional custom `id` provided as a prop to the hook.
 - A dynamically rendered `aria-expanded` attribute to let a screen reader know
   whether the disclosure component is visible.
-- The `aria-controls` attribute to let a screen reader know which disclosure
-  component is controlled by the button.
+- The `aria-controls` attribute using the custom `id` provided as a prop to the
+  hook. This lets a screen reader know which disclosure component is controlled
+  by the button.
 - An onClick handler that uses the `onToggle` callback along with any other
   click events passed as an `onClick` prop to `getButtonProps`;
 
--- a/pages/docs/styled-system/utility-hooks/use-disclosure.mdx
+++ b/pages/docs/styled-system/utility-hooks/use-disclosure.mdx
@@ -74,6 +74,13 @@ function Basic() {
 }
 ```
 
+If you need more control over the disclosure behavior, you can use a combination
+of the getter props along with the other methods and values returned by the
+hook.
+
+Below is use of the return fields of the hook without a getter to control the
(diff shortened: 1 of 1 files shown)


pages/docs/styled-system/utility-hooks/use-disclosure.mdx
```diff
@@ -1,60 +1,113 @@
...
+          <DrawerBody>
+            <p>Some contents...</p>
+            <p>Some contents...</p>
+            <p>Some contents...</p>
+          </DrawerBody>
+        </DrawerContent>
+      </Drawer>
+    </>
+  )
+}
```
noobinthisgame, line 105: I'd move this above the other example. This is the more basic way to use useDisclosure. The example you added is more advanced usage. Maybe  2 sub sections for usage could make sense: Basic and Advanced

noobinthisgame (changes requested):
Thanks for your work on this @TylerAPfledderer ! I've left a comment :)

pages/docs/styled-system/utility-hooks/use-disclosure.mdx
TylerAPfledderer: Oh! I figured because of only using the getters were the basic way because they do a simple toggle and handle needed accessibility, and you only specify a button for toggle and a component that is to be visibly toggled. I would like to understand why it is considered the advanced approach! 😄 

Aside from that, I will go ahead and make the change. 👍🏼

noobinthisgame: That's is just my opinion. I think so because in the more 'basic' usage you don't bother about anything except open/close something. We don't have to put it explicitly in sub sections, that might cause friction indeed.

TylerAPfledderer: Ok! I'll just flip the examples and tweak the wording of the details.

TylerAPfledderer: @noobinthisgame good to check this again. 😃

PR #594, opened 2022-05-12
docs(advanced): create sandpack for single part component section

TylerAPfledderer:
## 📝 Description

Provide an interactive example for the [Single Part Component](https://chakra-ui.com/docs/styled-system/theming/advanced#single-part-component) section using `Sandpack`.

The current code example does not have a live preview for the user to be able to practice manipulating the custom theme while applying the parts to a component.

## 🚀 New behavior

Because the code structure is not conducive to be rendered properly in the `ReactPreview` component, use `Sandpack` instead.

And provide a more robust version of the original example.

No


configs/sandpack-contents/single-component-style/custom-badge.js
```diff
@@ -0,0 +1,72 @@
...
+const rootElement = document.getElementById("root");
+ReactDOM.render(
+  <>
+    <ChakraProvider theme={customTheme}>
+    <ColorModeScript />
+    <App />
+    </ChakraProvider>
+  </>, 
+  rootElement
+);
```
noobinthisgame, line 70: Please update this to React 18 syntax. Also I would prefer to have this file as a .ts file, as in the tutorial for example.

noobinthisgame (changes requested):

noobinthisgame:
@TylerAPfledderer this looks good :) I've left a comment. Make sure you forward the changes from main into your branch.

configs/sandpack-contents/single-component-style/custom-badge.js
TylerAPfledderer: @noobinthisgame right. 😅 

While we are on the subject, at some point should the other `.js` files in `sandpack-contents` be converted to `.ts` as well?

noobinthisgame (approved):
LGTM! 🚀

PR #602, opened 2022-05-12
fix(changelog): turn off live previews to affected jsx examples

TylerAPfledderer:
## 📝 Description

JSX examples from the most recent changelogs in the [Changelog](https://chakra-ui.com/changelog) doc page are attempting to render preview, and end up breaking with visible error messages on the page.

Added `live=false` to these examples in the `changelog/index.mdx` file.


noobinthisgame (approved):
Thanks for fixing this @TylerAPfledderer !

PR #611, opened 2022-05-16
docs(popover): mention `forwardRef` with custom components

### LINKED ISSUE #513: Popover improvements - Components need ForwardRef!
### Subject

Popover, PopoverAnchor, PopoverTrigger

### Description

Current popover documentation for Anchor and Trigger do not have any mention of the React element being passed in needing a Ref, and that any custom component should work with ForwardRef. In retrospect, this is kind of a no-brainer, but if I wasted several hours debugging, then it would help someone else to have a kind reminder that passing in a custom component to either of these two wrapper elements requires a ForwardRef.

`PopoverTrigger` (emphasis added to highlight the one tiny section)
> Basic Usage[#](https://chakra-ui.com/docs/components/overlay/popover#basic-usage)
> When using this component, ensure the children passed to PopoverTrigger is focusable. Users can tab to it using their keyboard, **_and it can take a `ref`_**. It is critical for accessiblity.

Only barely mentions the "it can take a `ref`", and the way it's worded sounds like it's optional. 

"Ensure the children passed to PopoverTrigger is focusable and can take a ref, which is required for Popover to position correctly..." (as an example) is more direct. Without the `ref` the trigger _will not work_ to position the popover content correctly.

`PopoverAnchor` doesn't mention the ref at all.
> Popover Anchor[#](https://chakra-ui.com/docs/components/overlay/popover#popover-anchor)
> You can wrap your component with PopoverAnchor to prevent trigger any action. The wrapped component will become a position reference. Actions will only be triggerred by components inside PopoverTrigger.
> 
> In this case, you can only open and close the popover with Button. If you click on Input, it acts same as the original input and doesn't trigger any action about popover.

TylerAPfledderer:
Closes #513 

## 📝 Description

Add details in the [Basic Usage](https://chakra-ui.com/docs/components/overlay/popover#basic-usage) and [Popover Anchor](https://chakra-ui.com/docs/components/overlay/popover#popover-anchor) sections about using Chakra's `forwardRef` in custom components that are used as children in `PopoverAnchor` and `PopoverTrigger`.

### DIFF
--- a/pages/docs/components/overlay/popover.mdx
+++ b/pages/docs/components/overlay/popover.mdx
@@ -51,7 +51,7 @@ When using this component, ensure the children passed to `PopoverTrigger` is
 focusable and has a forwarded ref. Users can tab to it using their keyboard and
 is critical for accessibility. If you are using a custom component inside
 `PopoverTrigger` (or [PopoverAnchor](#popover-anchor)) it needs Chakra's
-[forwardRef](https://chakra-ui.com/docs/components/recipes/as-prop#option-1-using-forwardref-from-chakra-uireact).
+[forwardRef](../recipes/as-prop#option-1-using-forwardref-from-chakra-uireact).
 
 > **A11y:** When Popover opens, focus is sent to `PopoverContent`. When it
 > closes, focus is returned to the trigger.
@@ -290,7 +290,7 @@ action about popover.
 
 > 🚨 Note: As with [PopoverTrigger](#basic-usage), if you use a custom component
 > inside `PopoverAnchor` ensure that it uses Chakra's
-> [forwardRef](https://chakra-ui.com/docs/components/recipes/as-prop#option-1-using-forwardref-from-chakra-uireact)
+> [forwardRef](../recipes/as-prop#option-1-using-forwardref-from-chakra-uireact)
 > so the popover is triggered successfully.
 
 ```jsx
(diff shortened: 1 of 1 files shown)


pages/docs/components/overlay/popover.mdx
```diff
@@ -48,8 +48,10 @@ import {
 ## Basic Usage
 
 When using this component, ensure the children passed to `PopoverTrigger` is
-focusable. Users can tab to it using their keyboard, and it can take a `ref`. It
-is critical for accessiblity.
+focusable and has a forwarded ref. Users can tab to it using their keyboard and
+is critical for accessibility. If you are using a custom component inside
+`PopoverTrigger` (or [PopoverAnchor](#popover-anchor)) it needs Chakra's
+[forwardRef](https://chakra-ui.com/docs/components/recipes/as-prop#option-1-using-forwardref-from-chakra-uireact).
```
nikolovlazar, line 54:
```suggestion
[forwardRef](../recipes/as-prop#option-1-using-forwardref-from-chakra-uireact).
```

I think it's better to use a relative URL here instead of an absolute one, because we use subdomains like v0, v1, next, and also preview deployments as well.

nikolovlazar (changes requested):
Everything LGTM! Except that small comment I left below.

TylerAPfledderer:
@nikolovlazar Done. Should be good to go!

nikolovlazar (approved):
Alright, looking good!

PR #614, opened 2022-05-16
docs: update sandpack examples

TylerAPfledderer:
## 📝 Description

Update all Sandpack code examples in the docsite with the following changes:

- Convert code to React 18 API
- At least one example has its code inline in the `files` prop of the `SandPackEmbed` component. Move this code to the `sandpack-content` directory to de-clutter the doc file and provide consistency with other sandpack content.

### Examples to Update
- [x] [Floating Labels Recipe](https://chakra-ui.com/docs/components/recipes/floating-labels)
- [x] "Less code. More speed." section of the [Homepage](https://chakra-ui.com)
- [x] All examples, [ChakraUI + Formik](https://chakra-ui.com/guides/integrations/with-formik)
- [x] All examples, [ChakraUI + Framer Motion](https://chakra-ui.com/guides/integrations/with-framer)
- [x] [Advanced Formik Integration](https://chakra-ui.com/guides/recipes/advanced-formik-integration) 
   - Only addresses typing error related to ChakraUI Issue [#6023](https://github.com/chakra-ui/chakra-ui/issues/6023) as it is a quick fix. Other typing errors in the example should be addressed in a separate PR, if need be. (Example may need a revamp, typescript-wise?)
- [x] [ChakraUI Prose](https://chakra-ui.com/guides/recipes/prose)

## 📝 Additional Information
(VERY MINOR) Per the React 18 Concurrent Renderer, because the generated files in the `SandPackEmbed`s are `.tsx`, the container passed to `createRoot` needs a postfix to prevent an "is null" type error when a user views the example in CSB. (No problem removing this small change if requested! 😄)

### DIFF
--- a/configs/sandpack-contents/advanced-formik-integration/image-radio-buttons.js
+++ b/configs/sandpack-contents/advanced-formik-integration/image-radio-buttons.js
@@ -100,7 +100,7 @@ const Input = ({ name, ...props }: Props) => {
 
 export default Input;`,
 
-  App: `import { Box, ChakraProvider, Button } from "@chakra-ui/react";
+  App: `import { Box, Button } from "@chakra-ui/react";
 import { Formik, FormikProps } from "formik";
 
 import Input from "./input";
@@ -116,7 +116,7 @@ const AVATARS = [
 
 type Values = {
   email: string;
-  avatar: number;
+  avatar: string;
 };
 
 export default function App() {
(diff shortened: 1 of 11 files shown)


noobinthisgame (approved):
Thanks @TylerAPfledderer seems like I missed those when updating to v2 !

PR #622, opened 2022-05-19
fix(component-links): allow overflow of the container

### LINKED ISSUE #619: Focus Outline from `ComponentLink` clipped
### Description

In any component doc page containing the github/npm links to view source, theme, etc., on focus of the link the outline shadow is getting cropped off. This is due to the combination of the negative margin and `overflow: hidden` rendered by the `Wrap` component user used in the [`ComponentLinks` component](https://github.com/chakra-ui/chakra-ui-docs/blob/main/src/components/mdx-components/component-links.tsx).

The simplest and cleanest solution is to apply `overflow: visible` to the `Wrap` component used.

Side note: The hover states of each link are also impacted. Because the link shifts on hover, you can visibly see the border being clipped by the hidden overflow.

### Screenshot
image

TylerAPfledderer:
Closes #619

Revert the `overflow` prop originally set from the `Wrap` component that contains the component links (i.e. source code, theme, npm) to allow the boxShadow on focus and the transition on hover to not be clipped.

Clipping comes from the combination of `overflow: hidden` and negative margins, both set by the `Wrap` component.

### DIFF
--- a/src/components/mdx-components/component-links.tsx
+++ b/src/components/mdx-components/component-links.tsx
@@ -90,6 +90,7 @@ function ComponentLinks(props: ComponentLinksProps) {
     </ComponentLink>
   )
 
+  // Note: Currently an unused component
   const storybookLink = storybook?.url && (
     <ComponentLink
       url={storybook.url}
@@ -101,6 +102,7 @@ function ComponentLinks(props: ComponentLinksProps) {
     </ComponentLink>
   )
 
+  // Note: Currently an unused component
   const videoLink = video?.url && (
     <ComponentLink
       url={video.url}
--- a/src/components/mdx-components/component-links.tsx
+++ b/src/components/mdx-components/component-links.tsx
@@ -131,7 +131,13 @@ function ComponentLinks(props: ComponentLinksProps) {
   )
 
   return (
-    <Wrap className='component-links' mt='2rem' spacing='4' {...rest}>
+    <Wrap
+      className='component-links'
+      mt='2rem'
+      spacing='4'
+      overflow='unset'
+      {...rest}
+    >
       {githubLink}
       {themeComponentLink}
       {npmLink}
--- a/src/components/mdx-components/component-links.tsx
+++ b/src/components/mdx-components/component-links.tsx
@@ -21,7 +21,7 @@ type ComponentLinkProps = ButtonProps & {
 }
 
 function ComponentLink(props: ComponentLinkProps) {
-  const { icon, url, children, iconSize, iconColor, ...rest } = props
+  const { icon: BtnIcon, url, children, iconSize, iconColor, ...rest } = props
   return (
     <Button
       as={Link}
@@ -29,10 +29,12 @@ function ComponentLink(props: ComponentLinkProps) {
       isExternal
       px='12px'
       display='flex'
+      fontSize='sm'
       alignItems='center'
       minH='32px'
       borderWidth='1px'
       borderRadius='md'
+      marginInlineStart='0 !important'
       color={useColorModeValue('gray.600', 'whiteAlpha.700')}
       _hover={{
         color: useColorModeValue('gray.700', 'whiteAlpha.900'),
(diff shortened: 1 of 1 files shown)


src/components/mdx-components/component-links.tsx
```diff
@@ -1,144 +1,150 @@
...
+      spacing='4'
+      overflow='unset'
+      {...rest}
+    >
+      {githubLink}
+      {themeComponentLink}
+      {npmLink}
+      {storybookLink}
+      {videoLink}
+    </Wrap>
```
noobinthisgame, line 146: Could we replace this with a `ButtonGroup` and modify the `ComponentLink` to be a Button styled accordingly as Link ? I think that would clean up here this component a lot and be more chakra like

noobinthisgame (changes requested):
Nice catch @TylerAPfledderer. I've left some thoughts.

src/components/mdx-components/component-links.tsx
TylerAPfledderer: Ahh, yes! I'm on it! 🚀

TylerAPfledderer: @noobinthisgame I guess while I'm in here, do we still use the `ComponentLink`'s for Youtube and Storybook links? I've never seen them used in any of the docs.

noobinthisgame: Let's keep them for now, but add a comment above that they are unused if that's really the case.

TylerAPfledderer:
@noobinthisgame good to check again!

src/components/mdx-components/component-links.tsx
```diff
@@ -1,144 +1,145 @@
...
+      _hover={{
+        color: useColorModeValue('gray.700', 'whiteAlpha.900'),
+        bg: useColorModeValue('gray.300', 'whiteAlpha.300'),
+        boxShadow: 'sm',
+        transform: 'translateY(-1px)',
+        textDecor: 'none',
+      }}
+      {...rest}
+    >
+      <HStack>
```
noobinthisgame, line 46: You can use the leftIcon prop of the Button component to place the icons.

src/components/mdx-components/component-links.tsx
```diff
@@ -1,144 +1,145 @@
...
+      icon={FaGithub}
+      iconColor={iconColor}
+      iconSize='1rem'
+    >
+      {t('component.mdx-components.component-links.view-theme-source')}
+    </ComponentLink>
+  )
+
+  return (
+    <ButtonGroup
```
noobinthisgame, line 129: Here we have a little responsiveness issue when on phone.

src/components/mdx-components/component-links.tsx
```diff
@@ -1,144 +1,145 @@
...
+  icon: React.ElementType
+  url: string
+  iconSize?: string
+  iconColor?: string
+}
+
+function ComponentLink(props: ComponentLinkProps) {
+  const { icon, url, children, iconSize, iconColor, ...rest } = props
+  return (
+    <Button
```
noobinthisgame, line 26: It would be nice if the buttons could match the old design ? What do you think?

noobinthisgame (changes requested):
I've left some thoughts :)

src/components/mdx-components/component-links.tsx
TylerAPfledderer: With the Button component setup, have to use the `sx` prop to ensure the `iconSize` and `iconColor` passed in are used. No major issue there.

src/components/mdx-components/component-links.tsx
TylerAPfledderer: Sure! I personally favor retaining the touch space the `Button` provides for the mobile accessibility. There is a compromise here. 😄 
image

noobinthisgame: A compromise is what I meant. Was just missing the border and backgroundcolor when I checked :)

TylerAPfledderer:
@noobinthisgame alright, one more time! 😆

noobinthisgame (commented):
Good job ! :)

noobinthisgame (approved):

PR #624, opened 2022-05-19
fix(slider): update the first "Slider with custom labels and marks" example

TylerAPfledderer:
## 📝 Description

Update the slider example with the custom floating mark to ensure everything is visible in the live preview container.

Due to the live preview having `overflowX: auto`, the floating mark in it's absolute positioning is clipped out of view.

## 🚀 New behavior

To maintain integrity of the example, provide a wrapper `Box` component that sets padding to bring the floating mark into view.

This PR also tidies up the example with a small refactor.


noobinthisgame (approved):
LGTM 🚀  Thanks @TylerAPfledderer

PR #637, opened 2022-05-22
fix(get-search-meta): ignore tutorial pages on gen

TylerAPfledderer:
Currently, the `get-search-meta` script generates slugs for the incomplete tutorial pages. This PR add a Regex check for those file paths and ignores them by running an empty push to the object that gets written to the `search-meta.json` file.


scripts/get-search-meta.ts
```diff
@@ -88,11 +88,19 @@ async function getSearchMeta() {
...
-    try {
-      result = await getMDXMeta(file)
-      json.push(...result)
-    } catch (error) {
-      console.log(error)
+    /**
+     * * Ignore "tutorial" file paths with an empty push
+     * * Remove this check when tutorials are live!
+     */
+    if (/[\\/]pages[\\/]tutorial/gi.test(file)) {
```
segunadebayo, line 95: Could we simplify this to make it easier to ignore certain pages in the future. I wonder if we can do something like `const excluded = ['pages/tutorial']` and then omit based on items in that array?

Another idea is to update https://github.com/chakra-ui/chakra-ui-docs/blob/main/contentlayer.config.ts#L116 and add an `excludeInSearch: true` to the `computedFields`, and then filter the files based on that.

TylerAPfledderer: @segunadebayo I am not familiar with `contentlayer` but I have been working to understand it as I go! Based on what you suggested, it does not appear that adding `excludeInSearch` would work as expected, either because it is not an option in the config (type error) or it would not reach a component or script that would prevent the slugs from being added to `search-meta.json` (via frontmatter).

Since you suggested adding an array to check against, maybe this:
```js
/**
 * File paths to not be included in the search meta.
 *
 * This can be overall page sections (i.e. "/docs", "/tutorial", etc.) or specific files (i.e. "/guides/first-steps").
 */
const excludedSlugs = ['/tutorial']

for (const file of files) {
  let result: any[] = []

  // Windows OS: ensure file paths have forward slashes.
  const fileToPosixPath = posixPath(file)

  const isExcluded = excludedSlugs.find((excludedSlug) =>
    fileToPosixPath.includes(excludedSlug),
  )

  try {
    result = isExcluded ? null : await getMDXMeta(file)
    json.push(...result)
  } catch (error) {
    console.log(error)
  }
}
```

noobinthisgame:
@segunadebayo I'll leave this to you for final review :)

segunadebayo:
Let's resolve this conflict @TylerAPfledderer and I'll get this merged.

TylerAPfledderer:
@segunadebayo conflict resolved! 😄 ⚔️

PR #654, opened 2022-05-26
docs: use correct google font preconnect link

lwhiteley:
The preconnect link for google fonts in the Remix guide has a typo. This change corrects the incorrect url


nikolovlazar:
Good catch! Thanks, @lwhiteley!

Issue #669, opened 2022-05-30
Spelling mistake in CheckboxGroup


tanerijun:
### Description

The name `Kakashi` in the example code in [CheckboxGroup doc](https://chakra-ui.com/docs/components/form/checkbox#checkboxgroup) is not properly capitalized

m4x3d:
Fixed it with another open PR of mine.

![image](https://user-images.githubusercontent.com/44474134/170997448-dd55c637-c730-482c-9f5d-b9dc5971636c.png)

PR #670, opened 2022-05-30
docs: extend doc for size prop of heading

### LINKED ISSUE #667: How is "If you use the `size` prop, the font size of the heading will automatically decrease in size for smaller screens." true?
### Description

While reading through the documentation, I noticed that the [`<Text>` component examples uses `fontSize`](https://github.com/chakra-ui/chakra-ui-docs/blob/main/pages/docs/components/typography/text.mdx#changing-the-font-size) as props, while [`<Heading>` examples uses `size`](https://github.com/chakra-ui/chakra-ui-docs/blob/main/pages/docs/components/typography/heading.mdx#changing-font-size). 

The documentation explains this with
> To increase the size of the heading, you can use the fontSize or size props. **If you use the size prop, the font size of the heading will automatically decrease in size for smaller screens.**

However, how is this true if both `size` and `fontSize` props use rem as sizes?
- https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/foundations/sizes.ts
- https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/foundations/typography.ts#L47

Shouldn't the `size` prop then use an [array of values](https://chakra-ui.com/docs/styled-system/features/responsive-styles#the-array-syntax)?

Or why even have the inconsistency of size and fontSize. Stay true to the native attribute `fontSize` like you also mention [here in the documentation](https://chakra-ui.com/docs/styled-system/theming/theme#typography).

m4x3d:
Closes #667 and #669

## 📝 Description

Extension of documentation for breakpoints on the fontSize of `Heading` by using the `size` prop.

No

### DIFF
--- a/pages/docs/components/form/checkbox.mdx
+++ b/pages/docs/components/form/checkbox.mdx
@@ -188,7 +188,7 @@ shared styling props.
   <Stack spacing={[1, 5]} direction={['column', 'row']}>
     <Checkbox value='naruto'>Naruto</Checkbox>
     <Checkbox value='sasuke'>Sasuke</Checkbox>
-    <Checkbox value='kakashi'>kakashi</Checkbox>
+    <Checkbox value='kakashi'>Kakashi</Checkbox>
   </Stack>
 </CheckboxGroup>
 ```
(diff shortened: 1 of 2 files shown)


nikolovlazar (approved):
👌

PR #675, opened 2022-05-31
docs(text): make noOfLines prop consistent

### LINKED ISSUE #674: docs(text): Make noOfLines prop consistent
### Subject

Text

### Description

The docs currently have a redundant section regarding the `noOfLines` prop (probably due to the `isTruncated` to `noOfLines` change). Removing this extra section can make the docs clearer.

trybick:
Closes #674 

## 📝 Description

> Remove redundant section to make docs clearer

### DIFF
--- a/pages/docs/components/typography/heading.mdx
+++ b/pages/docs/components/typography/heading.mdx
@@ -67,8 +67,9 @@ the default theme. This can be seen in the [theme source](https://github.com/cha
 
 ### Truncate heading
 
-Pass the `noOfLines={1}` prop to render an ellipsis when the heading exceeds the
-width of the viewport or `maxWidth` prop.
+If you'd like to truncate the heading after a specific number of lines, pass the
+`noOfLines` prop. This will render an ellipsis when the heading exceeds the width
+of the viewport or `maxWidth` prop.
 
 ```jsx
 <Heading noOfLines={1}>
--- a/pages/docs/components/typography/text.mdx
+++ b/pages/docs/components/typography/text.mdx
@@ -41,32 +41,14 @@ To increase the font size of the text, you can pass the `fontSize` prop.
 
 ### Truncate text
 
-Pass the `noOfLines={0}` prop to render an ellipsis when the text exceeds the
-width of the viewport or `maxWidth` prop.
+If you'd like to truncate the text after a specific number of lines, pass the
+`noOfLines` prop. This will render an ellipsis when the text exceeds the width
+of the viewport or `maxWidth` prop.
 
 ```jsx
 <Text color='gray.500' noOfLines={1}>
   Lorem ipsum is placeholder text commonly used in the graphic, print, and
   publishing industries for previewing layouts and visual mockups.
-</Text>
-```
-
-Likewise, if you'd like to truncate the text after a specific number of lines,
-pass the `noOfLines` prop and set it to the desired no of lines.
-
-```jsx live=false
-// Basic version
-<Text noOfLines={2}>
-  "The quick brown fox jumps over the lazy dog" is an English-language pangram—a
-  sentence that contains all of the letters of the English alphabet. Owing to
-  its existence, Chakra was created.
-</Text>
-
-// Responsive version
-<Text noOfLines={[1, 2, 3]}>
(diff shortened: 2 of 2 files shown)


pages/docs/components/typography/text.mdx
```diff
@@ -51,25 +52,6 @@ width of the viewport or `maxWidth` prop.
 </Text>
 ```
 
-Likewise, if you'd like to truncate the text after a specific number of lines,
-pass the `noOfLines` prop and set it to the desired no of lines.
```
trybick, line 55: This "Likewise..." section is a little redundant because we already introduce the `noOfLines` above. 

If we look at the Heading docs, it does not have this section for `noOfLines`

pages/docs/components/typography/text.mdx
```diff
@@ -41,8 +41,9 @@ To increase the font size of the text, you can pass the `fontSize` prop.
 
 ### Truncate text
 
-Pass the `noOfLines={0}` prop to render an ellipsis when the text exceeds the
-width of the viewport or `maxWidth` prop.
+If you'd like to truncate the text after a specific number of lines, pass the
+`noOfLines` prop. This will render an ellipsis when the text exceeds the width
```
trybick, line 45: Changed `noOfLines={0}` to `noOfLines` here

pages/docs/components/typography/heading.mdx
```diff
@@ -67,8 +67,9 @@ the default theme. This can be seen in the [theme source](https://github.com/cha
 
 ### Truncate heading
 
-Pass the `noOfLines={1}` prop to render an ellipsis when the heading exceeds the
-width of the viewport or `maxWidth` prop.
+If you'd like to truncate the heading after a specific number of lines, pass the
+`noOfLines` prop. This will render an ellipsis when the heading exceeds the width
+of the viewport or `maxWidth` prop.
```
trybick, line 72: Also updated Heading docs to be more clear that `noOfLines` accepts any number, not just `noOfLines={1}`

noobinthisgame, line 72:
```suggestion
of the viewport.
```

noobinthisgame: This is implicit in my opinion.

pages/docs/components/typography/text.mdx
```diff
@@ -41,8 +41,9 @@ To increase the font size of the text, you can pass the `fontSize` prop.
...
+of the viewport or `maxWidth` prop.
```
noobinthisgame, line 46:
```suggestion
of the viewport.
```

See above.

pages/docs/components/typography/text.mdx
noobinthisgame: Good catch! I'd love to keep the responsive example though. Maybe you can add it to the example above. :)

noobinthisgame (changes requested):
Hey @trybick, thanks for working on this! I've left some thoughts

pages/docs/components/typography/text.mdx
trybick: @noobinthisgame Sure, I have updated the example to include the responsive version too. If this looks like a good example, I can also add this to the `Heading` component if we'd like since they both have a similar Truncate Text section.

pages/docs/components/typography/heading.mdx
trybick: Hmmm, I would say including this might be helpful to people. When I think of viewport I think of the whole screen (maybe that's just me). Including the maxWidth prop suggestion here could help some people get the truncation working faster. Thoughts?

noobinthisgame: Then lets leave it like this :)

pages/docs/components/typography/text.mdx
noobinthisgame: Thanks! I think that it is enough if we have it in one place. We can add it later if needed :)

noobinthisgame (approved):
It's ready to go :)

trybick:
@noobinthisgame Thanks for the review!

PR #677, opened 2022-05-31
docs(theme-tools): create doc page for the Theme Tools

### LINKED ISSUE #206: Add theme-tools to documentation
### Subject

Switch

### Description

We have noticed that there is a need to use the `cssVar` export from `@chakra/theme-tools` to properly style the Switch component size. This is not a part of the official framework documentation for component style.

If using `cssVar` for styling Chakra components is a public part of the theme framework and not an implementation detail, it should be documented in [CSS variables](https://chakra-ui.com/docs/features/css-variables) and in the components that require its usage.

[Discord conversation](https://discord.com/channels/660863154703695893/695564068895064094/928409199841525780)
[Codesandbox example](https://codesandbox.io/s/reverent-christian-hx716?file=/src/Theme.tsx)

Thank you for your consideration!

TylerAPfledderer:
Closes #206 

Add documentation to provide an overview of all the the Theme Tool methods that are publicly available for use in custom theme configs. Does not include the code for the anatomy.

## Considerations

Due to the repetitiveness of these methods (as they are mostly primitives in the arguments past and the values returned), it seems better to provide tables listing the methods, a description of what each one does and the value returned, and the basic function call to show what params are asked for in what order. This on it's face can certainly keep the clutter low, and be able to follow up each table with a set of examples from the Chakra source code.

Is creating a big table the best way to go here? This seems to be only needed for the color tools, but I think it's better than isolating each method to it's own section, since most good examples to provide from the source code show a mix of the methods to achieve certain dynamic styling results ([`getBg()` function](https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/alert.ts#L36-L41) from the `Alert` theme, for example)


pages/docs/styled-system/theming/theme-tools.mdx
```diff
@@ -0,0 +1,233 @@
...
+theme.
+
+| Name                 | Description                                                                                                                                                                                                                                                                                          | Function Call                                                         |
+| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------- |
+| `getColor`           | Retrieves a color hex value. Needs the theme and color name, and can take a fallback hex if the color name does not return a valid hex from the theme.                                                                                                                                               | `getColor(theme, color, fallback)`                                    |
```
noobinthisgame, line 79: After reading the result in the preview multiple time, I find this table hard to read and understand. What about just doing header with the function name and sections with the informations as you did for mode()

pages/docs/styled-system/theming/theme-tools.mdx
```diff
@@ -0,0 +1,233 @@
...
+    spinner: {
+      color: mode(`${c}.500`, `${c}.200`)(props),
+    },
+  }
+}
+```
+
+## Component Tools
+
+There are mny times where you need the theme to be reliant on the color mode
```
noobinthisgame, line 126:
```suggestion
There are many times where you need the theme to be reliant on the color mode

```

noobinthisgame (changes requested):
Hey @TylerAPfledderer sorry it took so long for a review! This is quite a lot :D Thanks for working on this. I've left some feedback. 

Also @nikolovlazar I'd love a second pair of 👀 and 🧠

pages/docs/styled-system/theming/theme-tools.mdx
noobinthisgame: Yeah you can add a note :) I commented it in the actual code but maybe two places would help

pages/docs/styled-system/theming/theme-tools.mdx
```diff
@@ -0,0 +1,233 @@
...
+| `getColor`           | Retrieves a color hex value. Needs the theme and color name, and can take a fallback hex if the color name does not return a valid hex from the theme.                                                                                                                                               | `getColor(theme, color, fallback)`                                    |
+| `tone`               | Determines if the given color is relatively "light" or "dark". Requires the theme and a color                                                                                                                                                                                                        | `tone(color)(theme)`                                                  |
+| `isDark` / `isLight` | Utilizes the `tone` method to return a boolean whether a color is "Dark" or "Light".                                                                                                                                                                                                                 | `isDark(color)(theme)`                                                |
```
nikolovlazar, line 81: All of the `color` methods from the `theme-tools` package are deprecated and will be removed: https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme-tools/src/color.ts#L40 maybe @segunadebayo or @TimKolberger can chime in here to let us know when exactly that'll happen. But regardless, since we've deprecated these there's no need to document them.

pages/docs/styled-system/theming/theme-tools.mdx
```diff
@@ -0,0 +1,233 @@
...
+use to simplify your custom theme file or styling within components using
+methods to make your design more dynamic.
+
+<ComponentLinks
+  github={{ package: 'theme-tools' }}
+  npm={{ package: '@chakra-ui/theme-tools' }}
+/>
+
+With any of these tools, you will need to install the `theme-tools` package
+separately as they are not a part of `@chakra-ui/react`.
```
nikolovlazar, line 16: I don't think the `theme-tools` package needs to be separately installed. I tried it quickly in CodeSandbox and I was able to pull in methods from that package without having to manually install it. Haven't tried to do that in a local VSC environment yet. Is that right, @segunadebayo?

nikolovlazar (changes requested):

segunadebayo:
We're going to deprecate this package in the next major. Let's close this out.

Thanks for your work on this Tyler

PR #717, opened 2022-06-13
style(props-table): add spacing between prop descriptions

### LINKED ISSUE #709: Update props table below components for better readability
### Subject

Text

### Description

I was trying to read through the props table and it's slightly confusing. There's technically nothing wrong, but i think it can be slightly improved! If you look at the image below, it looks like the documentation for `colorScheme` belongs to that section between the two borders, however, the documentation is actually before that section after the border. Is it possible to group the `prop` and description between the two borders for more clarity?

Thanks again! Love all the work you all have put in!

Screen Shot 2022-06-09 at 12 44 01 PM

TylerAPfledderer:
Closes #709 

## 📝 Description

Update spacing between props in the `PropsTable` so that the name of each prop did not appear as if it was a part of the previous prop description.

Current table (light mode):
image

## 🚀 New behavior

Updated table: (light mode):
image

### DIFF
--- a/src/components/props-table.tsx
+++ b/src/components/props-table.tsx
@@ -35,13 +35,7 @@ export type PropsTableProps = {
 
 const PropsTable = ({
   of,
-  omit = [
-    'layerStyle',
-    'noOfLines',
-    'textStyle',
-    'orientation',
-    'styleConfig',
-  ],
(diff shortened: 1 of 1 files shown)


mannycarrera4:
Thanks for the quick fix!

PR #727, opened 2022-06-16
fix(componenet-sidebar): change the order of items as it appears

eliyabar:
Closes nothing

## 📝 Description

This fix reorders the links of the components as presented in the docs.

When navigating from MEDIA AND ICONS section, from the Image component to the next component, you get to Transition, which isn't the first component in the OTHER section.

image

so the current NEXT order is:
Image-> Transitions -> Portal -> Show/Hide -> Close Button

## 🚀 New behavior

Fix the order to be sequential.
new NEXT order is:
Image -> Close Button -> Portal -> Show/Hide -> Transitions
Just like in the visual presented order
image

NO

## 📝 Additional Information

NO

### DIFF
--- a/configs/components-sidebar.ts
+++ b/configs/components-sidebar.ts
@@ -314,8 +314,8 @@ const sidebar: Routes = {
           summarize: true,
           routes: [
             {
-              title: 'Transitions',
-              path: '/docs/components/other/transitions',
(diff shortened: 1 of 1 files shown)


TylerAPfledderer:
LGTM. Nice catch, @eliyabar!

PR #753, opened 2022-06-23
Update Getting Started page to remove potential infinite loop

### LINKED ISSUE #751: Starter code on Getting Started page causes infinite loop
## Description
The following section of the Getting Started page will cause an infinite loop if used directly. The App function definition will be called recursively. While this is pretty easy to identify for most experienced developers, some of my students had issues with this.

https://github.com/chakra-ui/chakra-ui-docs/blob/fada8f3ef350a488edb205ebcfd99e42a0c4bf38/content/getting-started/index.mdx#L27-L34

## Resolution
There are few ways to resolve this, so I wanted to discuss before submitting a PR. The way I see it, we can either update the name of the function definition, or use a fake component inside the `ChakraProvider` to indicate a typical app container.

Another solution would be to mimic what a typical `index.js` file might look like instead of `App.js`, though the change from React 17 to 18 changed a few things relating to the root element, so a little more hesitant to suggest that. Maybe I'm over thinking and most people will be on v18 when adding Chakra for the first time anyway.

lachieh:
Closes #751 

## 📝 Description

The code sample on the Getting Started page will cause an infinite loop if used directly. The App function definition will be called recursively.

## 🚀 New behavior

New code sample does not call the App function recursively.

- [ ] Yes
- [x] No

### DIFF
--- a/content/getting-started/index.mdx
+++ b/content/getting-started/index.mdx
@@ -28,7 +28,7 @@ function App() {
   // 2. Wrap ChakraProvider at the root of your app
   return (
     <ChakraProvider>
-      <App />
+      <TheRestOfYourApplication />
     </ChakraProvider>
   )
 }


nikolovlazar (approved):
This would definitely solve the problem of copy-paste render loop. The `<TheRestOfYourApplication />` will throw an error because it's not imported, so the developers definitely won't miss it. Thanks for opening this PR, @lachieh 🙏

PR #772, opened 2022-06-25
build(deps): remove babel and superjson

### LINKED ISSUE #582: Use swc instead of babel
### Description

Check if we can replace babel with swc to build our docs page (the default for nextjs)

Before doing so it has to be checked:
- Is emotion supported 
- Do we need superjson
- Are there other breaking things

TylerAPfledderer:
Closes #582

Remove `superjson` and `babel` in favor of `swc`

Removal of the dependencies and rebuilding locally of the docsite yielded no errors, including navigating the site via spinup of the local production server.

> Posting PR Draft to also let Vercel run build and check out preview that way.


TylerAPfledderer:
@noobinthisgame circled back to this. Removal of these dependencies seemed to have `0` effect on the build. Doc pages are running as expected.

noobinthisgame:
That is what I expected, as the content for our pages is generated and provided by contentlayer.

noobinthisgame (approved):
Thanks for getting that done @TylerAPfledderer :) 🚀

PR #782, opened 2022-06-27
Update color-mode.mdx

yagudaev:
Closes: N/A

## 📝 Description

Updates documentation to use the proper path to `initialColorMode`. The theme object doesn't have a `config` key, it just contains all the keys.

Perhaps this is left over from a previous version/refactoring.

Using example will return `undefined` and fallback to using a `light` color mode.

## 🚀 New behavior

Create the intended behavior of using the system color scheme and allowing the user to override it with a button.

No

### DIFF
--- a/content/docs/styled-system/color-mode.mdx
+++ b/content/docs/styled-system/color-mode.mdx
@@ -140,7 +140,7 @@ export default class Document extends NextDocument {
         <Head />
         <body>
           {/* 👇 Here's the script */}
-          <ColorModeScript initialColorMode={theme.config.initialColorMode} />
+          <ColorModeScript initialColorMode={theme.initialColorMode} />
           <Main />
           <NextScript />
         </body>


noobinthisgame:
Hey @yagudaev, I've double-checked and came to the result that the page doesn't need an update. I've put together an example in this [codesandbox](https://codesandbox.io/s/wizardly-shtern-3dd6zm?file=/src/index.tsx) following our docs.

yagudaev:
Thanks for the quick reply @noobinthisgame. 

The example you gave only works because you use `{ config }` as a parameter to extendTheme. However, all the examples on the page use a plain object without first nesting it under a `config` key. This is where the confusion occurs 😁.

Here is an example: https://codesandbox.io/s/loving-forest-e28byd?file=/src/index.tsx.

Does that make sense?

noobinthisgame:
I see what you mean but the example in the current docs defines `config` as object and then it used as this: `extendTheme({ config })` which is the same as defining a `config` key. This is also what I did on in my code sandbox by using the example.

yagudaev:
Right, I first customized my design tokens and then went to add dark mode compatibility.

If you look at this documentation: https://chakra-ui.com/docs/styled-system/customize-theme, you will see why I did it this way.

Screen Shot 2022-07-04 at 5 58 25 PM

Are you saying lines 13-14 need encapsulation inside `config`, while the rest remain as-is?

If so, it would be good to have a `Theme` type and a `ThemeConfig` type in TS to avoid confusion.

noobinthisgame:
Yup that's what I meant :) Improving the type for the ThemeConfig is on our todo-list but is rather complex so we postponed it until someone takes up on the challenge. 

Feel free to close this PR if everything is clear :)

PR #785, opened 2022-06-28
Add more examples of different Input Methods

happycoder0011:
Closes # 

## 📝 Description

> Add a brief description
The documentation page just shows examples on text and password as input. It would be a great addition to show more examples such as date, month, time and color picker inputs. I saw reddits and stackoverflow questions asking for date and time picker exclusively not being available in Chakra UI as Material UI and ANTD.

No Mention of any other input methods supported by Chakra UI Input. Users expect exclusive date time pickers similar to Material UI and ANTD.
## 🚀 New behavior
Added examples on other input methods for the users.

Somewhat - need to install 3rd party libraries for specific components.

### DIFF
--- a/content/docs/components/input/usage.mdx
+++ b/content/docs/components/input/usage.mdx
@@ -217,3 +217,46 @@ opacity to 1 if you want the placeholder to have a specific color.
   />
 </Stack>
 ```
+
+### Input Methods other than Text
+Date and Time Picker
+
+```jsx
+  <Input
+              placeHolder="Select Date and Time"
+              size="md"
+              backgroundColor="#ffffff"
+              type="datetime-local"
+            />
+```
+Month Picker
+
+```jsx
+  <Input
+              placeHolder="Select Date and Time"
+              size="md"
+              backgroundColor="#ffffff"
+              type="month"
+            />
+```
+Time Picker
+
+```jsx
+  <Input
+              placeHolder="Select Date and Time"
+              size="md"
+              backgroundColor="#ffffff"
+              type="time"
+            />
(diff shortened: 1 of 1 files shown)


content/docs/components/input/usage.mdx
```diff
@@ -217,3 +217,46 @@ opacity to 1 if you want the placeholder to have a specific color.
...
+### Input Methods other than Text
+Date and Time Picker
+
+```jsx
+  <Input
+              placeHolder="Select Date and Time"
+              size="md"
+              backgroundColor="#ffffff"
+              type="datetime-local"
+            />
```
noobinthisgame, line 230: I'd go with only this example and add a note about more types and a link to [this list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#input_types)

Also please clean up the indentation

happycoder0011: ### Input Methods other than Text
You can use different types with `Input` such as dateTime, color, search, file etc. 

Date and Time Picker

```jsx
  <Input
     placeHolder="Select Date and Time"
     size="md"
     backgroundColor="#ffffff"
     type="datetime-local"
    />
```
Check for different input types available : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#input_types 

Does this looks good?  Or any other change suggestion?

noobinthisgame: Yes that's exactly what I meant :)

noobinthisgame (approved):
Thanks for adding this @happycoder0011

PR #791, opened 2022-07-01
refactor: replace react live with sandpack (WIP)

### LINKED ISSUE #790: use sandpack everywhere
### Subject

react-live => sandpack

### Description

I'd like for us to use the same package for code editing everywhere in the docs. The beta version of the React docs showcases the power of Sandpack (https://beta.reactjs.org/learn).

We need to leverage this in our docs as well.

TylerAPfledderer:
Closes #790

Replace all instances of the React Live Previews with the use of the CSB Sandpack .

> Note: Any instance of the word `Sandpack` will be abbreviated to `SP` throughout this overview.

As of 7/1: Separate component built to test the functionality before replacing the current `SandpackEmbed`.

### Implementations (Updated 7/7)
- Via a prop, allowed layout flexibility to account for both doc pages and the embed used on the homepage.
- Remove the `SandpackPreview` component and enable `SandpackCodeEditor`'s `readOnly` prop if an mdx code block contained `live=false`
- Included the `CopyButton` component that is used in the current examples, and obtain the code for copy via the `useActiveCode` hook from SP

#### Considerations
- Because most code examples use Chakra components, SP requires that they are imports in the generated file they are shown in. (React Preview did not require this).
  - Due to the level of variety in code examples presented, imports and export function blocks will need to be added within the mdx code blocks in order for the SP instance to work. Simplifying the process on the front-end by using a prop, say, will potentially add further complexity on the back-end to account for the variety of examples used. (Maybe this can be rectified at a later time).
  Example basic code block format requirement for live previews:
  ```jsx
  import { Button } from '@chakra-ui/react'

  export default function Example() {
    return <Button colorScheme='blue'>Button</Button>
  }
  ```
  Having to do the above seems somewhat consistent with how [beta.reactjs.org](https://beta.reactjs.org/learn) is using the embeds.
- Convert existing instances to use the new implementation (which should result in removing the related files from the `config` directory, as they would become code blocks directly in the mdx files)
- Prevent an embed instance from using local storage data set by any other instance. (This might not be possible)
  - i.e. Currently, if you toggle color mode in the [Advanced Theming](https://chakra-ui.com/docs/styled-system/advanced-theming) example, it will affect the background color rendered in all other instances
- Generate multiple files in a SP instance based on codeblocks that are written one after another, and possibly a prop to label them accordingly. (This is also an attempt to eliminate importing of config files with this data)

Based on the above points, there is the high likelihood that almost all code examples will need to be modified in some way. It might be helpful to include documentation in `CONTRIBUTING.md` to outline the basic requirements or suggestions in ensuring new or modified code examples are handled successfully.


TylerAPfledderer:
@noobinthisgame would you be able to take a look at this before I continue?

First off, I was trying to merge `main` into this branch, and I didn't expect all the commits to be visible in this PR vs Github just saying `Merge branch 'main' into feat/sandpack-all-docs` 🤣 

A few things to note when viewing:
1. This is a brand new component being used call `SandpackCodeBlock` found in `components` > `mdx-components`. This was to help with testing, by isolating the intended implementation and to plug it into individual code blocks to check results.
2. To save from polluting the commits, it is currently applied to the embed on the homepage and the [Style Props - Shadow](https://chakra-ui.com/docs/styled-system/style-props#shadow) section.
3. Any other embeds rendered aside from above are from the live site.
4. If you want to see what the component looks like in other places, you can go to a code block in any `mdx` file and add `type=sp-test` to the beginning of the block
```
```jsx type=sp-test
```

I currently do not know yet how this will work in the docs for the `Advanced Theming` or the recipes. I am intending to move the code from the config file directly into the `mdx` files. (Is this OK?)

I would love to know what your thoughts are. 😃

TylerAPfledderer:
Hey @nikolovlazar. If you have a moment....

I tried to clean up the commits, and somehow ended up adding more clutter in this PR. 🤦🏼 Do you believe that the file changes are still good, or do I need to fix this somehow? (Scrap this PR for a new one plus a new feature branch?)

The only files that are specific to this PR that I have committed are `mdx-components/mdx-components` and `mdx-components/sandpack-codeblock`

TylerAPfledderer:
Closing PR in favor of a new one due to unexpected issues with commits.

PR #803, opened 2022-07-04
fix(docs): removing line wrong redirecting to cra guide

nishu-murmu:
Closes # 

## 📝 Description
Removing the last line telling us if we want to generate cra template from scratch. The link is also wrong.

No


noobinthisgame (approved):
Good catch @nishu-murmu ! We missed that one when merging both pages.

PR #810, opened 2022-07-05
fix(docs): update tailwind dark mode information

Spikatrix:
## 📝 Description

Tailwind dark mode is [enabled by default](https://tailwindcss.com/docs/upgrade-guide#remove-dark-mode-configuration) from Tailwind v3 onwards. They use the new JIT compiler so file size is no longer a concern.

Docs mention that tailwind dark mode needs to be explicitly enabled

## 🚀 New behavior

Removed the part where the docs mention that tailwind dark mode needs to be explicitly enabled as this is no longer the case

No

## 📝 Additional Information

None


noobinthisgame (approved):
Thanks @Spikatrix!

PR #819, opened 2022-07-07
Split a non copyable codeblock into 2

theskinnycoder:
## 📝 Description

> Due to the **or** comment inside the codeblock to install the dependencies, the copy button copies both `npm` and `yarn` commands with the or comment in between. This PR split that uncopyable codeblock into two.

> The copy to clipboard button on the codeblock is redundant if comments and multiple alternate commands are getting copied.

## 🚀 New behavior

> Instead of 1 uncopyable codeblocks, this PR splits that into two, one for `yarn` users, the other for `npm` users.

> No


content/getting-started/index.mdx
```diff
@@ -8,9 +8,11 @@ To use Chakra UI in your project, run the following command in your terminal:
 
 ```bash
 npm i @chakra-ui/react @emotion/react @emotion/styled framer-motion
+```
 
-# or
+or
```
noobinthisgame, line 13: I'd remove the "or" and rephrase the sentence above to something like: `To use Chakra UI in your project, run one of the following commands in your terminal:`

noobinthisgame (changes requested):
@theskinnycoder good catch! I've left some thoughts :)

content/getting-started/index.mdx
```diff
@@ -8,9 +8,11 @@ To use Chakra UI in your project, run the following command in your terminal:
...
 
+```
```
TylerAPfledderer, line 15: I would also add that this second code block also needs the `bash` keyword so that this line for the yarn install is highlighted correctly.

` ```bash `

theskinnycoder: Thanks for the suggestion. Implemented it!

content/getting-started/index.mdx
theskinnycoder: Thanks for the suggestion! Made changes accordingly

theskinnycoder (commented):
Good point, considering this into the next commit

TylerAPfledderer (approved):
LGTM

TylerAPfledderer:
@all-contributors please add @theskinnycoder for doc

noobinthisgame:
@all-contributors please add @theskinnycoder for doc

PR #821, opened 2022-07-07
chore(FormControl): remove unnecessary htmlFor and id

LFerronato:
## 📝 Description

- Little changes on some examples, removing unnecessary `htmlFor` and `id`;
- And improvements about usage of this properties at `Improvements from v1` section.

Today, the [documentation](https://chakra-ui.com/docs/components/form-control/usage) of `FormControl` shows `htmlFor` and `id` as necessary props (it doesn't say that, but it does imply)

## 🚀 New behavior

Most samples will not show these props. However, their use is described for situations of greater need for control.

No

### DIFF
--- a/content/docs/components/form-control/usage.mdx
+++ b/content/docs/components/form-control/usage.mdx
@@ -121,9 +121,9 @@ indicate that a field is optional you can add `optionalIndicator` to the
 
 ```jsx
 <FormControl>
-  <FormLabel htmlFor='amount'>Amount</FormLabel>
+  <FormLabel>Amount</FormLabel>
   <NumberInput max={50} min={10}>
-    <NumberInputField id='amount' />
+    <NumberInputField />
     <NumberInputStepper>
       <NumberIncrementStepper />
       <NumberDecrementStepper />
--- a/content/docs/components/form-control/usage.mdx
+++ b/content/docs/components/form-control/usage.mdx
@@ -16,6 +16,7 @@ Chakra UI exports 4 components for Form Control:
   children.
 - **`FormLabel`**: The label of a form section. The usage is similar to
   [html label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label).
+  `htmlFor` is set by default for children input.
 - **`FormHelperText`**: The message that tells users more details about the form
   section.
 - **`FormErrorMessage`**: The message that shows up when an error occurs.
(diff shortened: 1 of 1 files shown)


LFerronato:
do you think it makes sense? thanks

TylerAPfledderer:
Hi @LFerronato! Thank you for your interest in improving the DocSite. 🌝

The use of `htmlFor` and the corresponding `id` in all the examples was originally added because it was believed to be a bug without them. (Clicking the label without `htmlFor` did not produce focus to the form element)

But on further inspection of the source code the form control provider is supplied a declared `id` variable with a random hash that is accessed by the label and any form element. The `htmlFor` with an `id` attribute is optional should the user need more control. I played around with the example code and checking out DevTools and everything seems to be working and rendering as intended without them.

Whether or not we are to remove this combo from the code examples, I think it would at least be a good idea to update the notes in the [Improvements from v1](https://chakra-ui.com/docs/components/form-control/usage#improvements-from-v1) section.

- You can still supply an `id` prop to `FormControl` that will override the randomly generated `id` and attach to the `for` attribute of the label and the `id` of the form element. (It also affects the `id` attribute of the label)
- The combination of `htmlFor` and `id` are optional in which adding both will also override the generated `id` sent to the provider.

Those are my thoughts if you think there is a better way to word it! :)

LFerronato:
Awesome @TylerAPfledderer, thanks for your validation on the source code and your suggestion!
I will edit, adding this note at the end.

Mainly, my intention with this PR, started when I was showing ChakraUI to my teammates, and I said "look how cool this FormGroup is, we don't need to define htmlFor in the label, because the FormGroup handles with that... [later besides, I opened doc about FormGroup...] oops!, here in the doc it is showing that we have to define `htmlFor` and `id`... I thought this was optional, let's test.. [than tested, e.g. ..haha] Yes! It's optional, maybe the documentation is out of date"

LFerronato:
I think that's it guys.
this is my first official contribution to an open source project.
And I love Chakra, and I hope it grows a lot!

TylerAPfledderer:
@LFerronato thanks for adding notes at the bottom.

For consistency sake, it looks like the example for the [Number Input](https://chakra-ui.com/docs/components/form-control/usage#number-input-example) section still needs to be updated. :)

LFerronato:
So I thought about this case. 
But I thought it would be better to use it since it has a lot of sub-elements. But, in reality, there is only one input, so the id will be defined in that input. I tested it here, and it's all right.
I'll commit this change.

TylerAPfledderer (approved):
No problem, @LFerronato! Makes sense.

LGTM! 🚀

LFerronato:
Hello guys, any update for this PR?
how can we proceed?

@TylerAPfledderer

TylerAPfledderer:
Sorry @LFerronato! Merging now. 😅

TylerAPfledderer:
@allcontributors please add @LFerronato for docs.

PR #879, opened 2022-07-31
Clarify the parameters for useBreakpointValue

oBusk:
## 📝 Description

The documentation doesn't really make it clear that you can define what breakpoint to use by default. I've just modified it to mention this.

### DIFF
--- a/content/docs/hooks/use-breakpoint-value.mdx
+++ b/content/docs/hooks/use-breakpoint-value.mdx
@@ -50,10 +50,3 @@ function Example() {
   )
 }
 ```
-
-## Parameters
-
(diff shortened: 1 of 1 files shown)


oBusk:
I noticed that the PropsTable seems to be written specifically for components and doesn't really work with this hook. Not sure if it would be ok to hardcode a table into the page to define the params? [Removed](https://github.com/chakra-ui/chakra-ui-docs/pull/879/commits/2c1b0c01437e2fa443a17eab7973729ac804de08) the parameters section completely for now.

noobinthisgame (approved):
Nice one @oBusk !

PR #959, opened 2022-08-30
Update style-props.mdx to match overflow props to correct CSS props

jonahallibone:
Closes # 

## 📝 Description

> Add a brief description

Fix `overflowX` and `overflowY` to match the correct `overflow-*` css props


nikolovlazar:
Good catch, @jonahallibone! Thanks for opening the PR 🙌

PR #962, opened 2022-09-01
Fixed `shouldForwardProp` example in the Framer Motion integration doc

Andarist:


configs/sandpack-contents/framer-motion-examples/chakra-factory.js
```diff
@@ -1,15 +1,14 @@
...
-  
+
 const ChakraBox = chakra(motion.div, {
   /**
-   * Allow motion props and the children prop to be forwarded.
-   * All other chakra props not matching the motion props will still be forwarded.
+   * Allow motion props and non-Chakra props to be forwarded.
    */
-  shouldForwardProp: (prop) => isValidMotionProp(prop) || prop === 'children',
+  shouldForwardProp: (prop) => isValidMotionProp(prop) || shouldForwardProp(prop),
```
Andarist, line 9: I'm not completely sure if this is the best way to implement this but it works - perhaps you would know how to correct it if needed. I assume that this is the `shouldForwardProp` that you are using by default with Emotion so it's probably the correct thing.

The problem I was facing is that I've used what was recommended in the docs and that prevented my `data-*` attributes from being forwarded to Framer Motion (and thus also prevented them from being forwarded to the DOM).

configs/sandpack-contents/framer-motion-examples/chakra-factory.js
```diff
@@ -1,15 +1,14 @@
 module.exports = {
-  App: `import { Container, chakra } from '@chakra-ui/react';
+  App: `import { Container, chakra, shouldForwardProp } from '@chakra-ui/react';
 import { motion, isValidMotionProp } from 'framer-motion';
-  
+
 const ChakraBox = chakra(motion.div, {
   /**
-   * Allow motion props and the children prop to be forwarded.
-   * All other chakra props not matching the motion props will still be forwarded.
```
Andarist, line 8: I've removed this second part of the JSDocs description because I found it confusing (and I think that it was incorrect with the code before my change as non-motion props were not forwarded at all)

nikolovlazar (approved):
Thanks for looking into this, @Andarist! LGTM!

PR #989, opened 2022-09-15
docs: update nextjs-guide

hwanyoungChoi:
Closes # 

## 📝 Description

> Modify chakra ui installation guide in nextjs

> Module not found: @chakra-ui/theme tried to access @chakra-ui/styled-system 
from yarn v3.2.3

## 🚀 New behavior

> Add package install item: @chakra-ui/styled-system

### DIFF
--- a/content/getting-started/nextjs-guide.mdx
+++ b/content/getting-started/nextjs-guide.mdx
@@ -28,11 +28,11 @@ category: frameworks
 In your Next.js project, install Chakra UI by running either of the following:
 
 ```bash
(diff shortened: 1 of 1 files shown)


noobinthisgame (commented):
As I cannot reproduce any issues with this installation command, when using npm, yarn classic or pnpm I'd leave this as it is. 

If this issue persists on your side, I'd be ok to add an Information for usage with yarn v3 but not to change the actual command.

nikolovlazar:
Totally agree with @noobinthisgame. We shouldn't change the command itself, but it's going to be okay if we provide a disclaimer below the command for people using Yarn V3. @hwanyoungChoi can you do that instead?

PR #1002, opened 2022-09-20
docs: remix guide color mode ssr enhancement

alexinfurs:
## 📝 Description

This PR improves Color Mode SSR with Remix.run

Currently, cookies are loaded within loader, so if you switch the theme and then do a client routing, it switches again to the previous theme
Also, the server doesn't include the relevant classes to html and body tags, so if you disable JavaScript the theme is not correctly rendered

## 🚀 New behavior

With this PR, cookies are overridden on the client side to avoid unwanted theme switch and the server adds classes to html and body tags to correctly render the theme without JavaScript enabled in the client.

No


content/getting-started/remix-guide.mdx
```diff
@@ -313,23 +314,90 @@ export const loader: LoaderFunction = async ({ request }) => {
...
-        }>
-        [...]
+  <html
+    lang="en"
+    {...(colorMode
+      ? {
+          "data-theme": colorMode,
+          "style": { colorScheme: colorMode },
+        }
+      : null)}
```
aacevski, line 363:
```suggestion
    {...colorMode
      && {
          "data-theme": colorMode,
          "style": { colorScheme: colorMode },
        }
      }
```

content/getting-started/remix-guide.mdx
```diff
@@ -313,23 +314,90 @@ export const loader: LoaderFunction = async ({ request }) => {
...
+          key={key}
+          data-emotion={`${key} ${ids.join(" ")}`}
+          dangerouslySetInnerHTML={{ __html: css }}
+        />
+      ))}
+    </head>
+    <body  
+      { ...(colorMode ? {
+        className: `chakra-ui-${colorMode}`
+      } : null )}
```
aacevski, line 379:
```suggestion
      {...colorMode && {
        className: `chakra-ui-${colorMode}`
      }}
```

aacevski (changes requested):
Hey @alexinfurs, sorry for leaving this PR stale for so long, this looks great. I just have one thing that we could add. Instead of the last part for community boilerplates, we could replace it with a CodeSandbox example and we can see how Chakra UI works with Remix in one rounded example.

aacevski:
One more question, why aren't we using hydrateRoot like in the official Remix guide? Maybe we should update.

alexinfurs:
I made just the minimal changes from the previous guide.
We definitely should update here, I already updated my projects to hydrateRoot and renderToPipeableStream to take advantages from React 18 streaming feature

aacevski (approved):

PR #1014, opened 2022-09-26
Added useDisclosure

robert-s-hogan:
Not sure if this is helpful.  I find myself forgetting to import useDisclosure with the rest of the modal components.  If this doesn't apply, no worries. 😄

Closes # 

## 📝 Description

> Add a brief description

## 🚀 New behavior


nikolovlazar (commented):
Sure thing @robert-s-hogan! We should also update the sentence above the list to reflect on the addition, something like

> Chakra UI exports 7 components to help you create any modal dialog. Additionally, you can use the `useDisclosure` hook to control when the modal shows/hides.

noobinthisgame (changes requested):
The actual import command needs an update as well to reflect that change :) Otherwise good Idea!

PR #1018, opened 2022-09-28
feat: add internalization

### LINKED ISSUE #475: docs(i18n): Add translation mechanism
### Description

Chinese documentation is so important ! 

### Problem Statement/Justification

There are a large number of developers in China who are limited by their English level and cannot use many advanced and powerful libraries.

### Proposed Solution or API

good

### Alternatives

_No response_

### Additional Information

_No response_

### Proposed workflow
- [ ] Check if localization works
- [ ] Check the language switcher
- [ ] Bring back the i18n config in Next.js
- [ ] Create a new LOCALIZATION.md document that outlines how to contribute to the localization
- [ ] Create a new folder in `i18n` to keep the alternative-language docs
- [ ] Modify the `getStaticProps`, get the `locale` property from the `ctx` and return the corresponding `mdx` file
- [ ] Figure out a way to notify translations when changes occur in the existing English pages

noobinthisgame:
Closes #475 

## 📝 Description

To address the need of internalization for our docs I figured out a way to provide multiple language support for all components and docs pages. I tried to keep the structure straight forward for contributors.

The content folder now splits up in the different locale sub folders e.g. "de", "en", etc. Additionally we have a `locales` folder, which has the same folder structure, with all strings we are using in our components and pages that are not in the content folder. (index.tsx, team.tsx for example)

Based on the local chosen by the user we have new application routes prefixed with the locale "/de/docs/components..." . Except for the "en" locale as it is the default one.

Only english was supported.

## 🚀 New behavior

Multiple languages can be added and toggled via a select on the homepage.

No breaking changes are introduced for users, as for the default locale no paths changed

## 📝 Additional Information

It would be nice to use a tool like crowdin to manager our translations. Otherwise it's going to be hard to keep an overview.

Open todos:

- [x] Enable locales for blog
- [x] Enable locales for changelog
- [x] Enable locales for community
- [x] Enable locales for getting-started
- [x] Enable locales for components
- [x] Polish language select
- [x] Refactor global locals and split up in files


noobinthisgame:
@nikolovlazar and @segunadebayo feel free to leave some thoughts here.

src/components/header.tsx
```diff
@@ -64,6 +65,11 @@ function HeaderContent() {
             flexShrink={0}
             display={{ base: 'none', md: 'flex' }}
           />
+          <LanguageSwitcher
+            width='auto'
+            flexShrink={0}
+            display={{ base: 'none', md: 'flex' }}
+          />
```
aacevski, line 72: To be fair I would style the language switcher to look a little better, It looks a little too small. 
image

This is what I found from searching on the web.

src/components/language-switcher.tsx
```diff
@@ -0,0 +1,39 @@
+import { Select, SelectProps } from '@chakra-ui/react'
+import { useRouter } from 'next/router'
+
+const languages = [
+  { label: '🇬🇧', value: 'en' },
+  { label: '🇩🇪', value: 'de' },
+]
```
aacevski, line 7: What if we have the whole names of the language instead instead of GB/DE? Also to me `EN` makes more sense since GB more reminds me of the country rather than the language.

aacevski (changes requested):
Since this PR is too big I will write down the issues I found while testing:
- Component showcase links are broken in German
- Searching components doesn't work in German

Other than that, everything worked okay for me. Kudos on the amazing work!

src/components/header.tsx
noobinthisgame: I will do so :) Glad you pointed it out. I was unsure whether to have text or icons.

src/components/language-switcher.tsx
noobinthisgame: The short form here is a convention for localization. It's the most convienient way to work with it.

noobinthisgame:
Good catches I'll have a look at those issues.

aacevski (changes requested):
Hey, great job man! All my comments have been resolved and I love the look! Just one small issue, when I'm on mobile the language picker doesn't seem to work, it just changes the color mode:
brave_1MjfYZrNjf

aacevski (approved):

segunadebayo:
Thanks for working on this @noobinthisgame.

I'm weary about the sheer amount of work required to get the translations done. A part of me feels like we're not ready for this, given the manpower we have.

I'd love to keep this solution, though, 'coz I like the approach. Given the other projects in the pipeline, I don't know if we need to have this right away.

Let's chat about this.

noobinthisgame:
@segunadebayo I feel the same about it. Reason why it is not merged yet. We can have a chat about it.

Glad you like this approach :)

For new projects having translating infrastructure early on would be beneficial (if we aim for translated content), as translation will be a community effort and take time.

noobinthisgame:
Closing this to keep it as reference

PR #1056, opened 2022-10-05
(docs) Next JS: Change custom document to use functional components.

TomRadford:
## 📝 Description

> Using a functional component over a class component is the recommended method to override the default document as per Next documentation: https://nextjs.org/docs/advanced-features/custom-document

> Docs for NextJS Custom Document use class component

## 🚀 New behavior

> Docs for NextJS Custom Document use functional component

No


noobinthisgame (approved):
Thanks for this change @TomRadford !

PR #1057, opened 2022-10-05
Update docs with defineStyleConfig et al

selbekk:
## 📝 Description

This change updates the docs for creating custom themed components, so that it uses the same tools as Chakra does internally - namely the `defineStyleConfig` and `createMultiStyleConfigHelpers` functions.

No


noobinthisgame (approved):
Nice one @selbekk :)

Issue #1041, opened 2022-10-03
Theming: Button


nikolovlazar:
### Subject

Button

### Description

Create a basic theming documentation for the `Button` component that showcases how you can style it.

Theming file: [Button theming file](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/button/theming.mdx)

dev-zaid:
Can you assign this task to me?

nikolovlazar:
Closing this because I'm providing the `Button` component theming docs as an example

PR #1065, opened 2022-10-09
Feat: updated button theming

deevo-sage:
Closes # Theming: Button #1041 

## 📝 Description

Added theming info for the button component

refered styles from [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/button.ts)

language similar to [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/select.ts)


nikolovlazar:
Hey @deevo-sage. I'll close this PR because I'm providing the Button component's theming docs as an example.

PR #1114, opened 2022-10-14
fix(toc-nav): redirect to main component page on link clicked

### LINKED ISSUE #1113: Anchor not working on props tabs
### Description

When I click on 
Capture d’écran 2022-10-14 à 16 57 25
On **props** tab, nothing happen except that the hash is added to the route. We don't change to go to **usage**
Capture d’écran 2022-10-14 à 17 00 14

Expected behavior : fallback to usage tab with right anchor and scroll to it.

P0ppoff:
Closes #1113

## 📝 Description

I add a ternary to reach usage tab with anchor.

Currently we stay on other tabs as **props** or **theming** but the anchor is added to the path.

## 🚀 New behavior

We navigate to **usage** tab with the correct anchor.

### DIFF
--- a/src/components/table-of-content.tsx
+++ b/src/components/table-of-content.tsx
@@ -10,6 +10,7 @@ import { useScrollSpy } from 'hooks/use-scrollspy'
 import { t } from 'utils/i18n'
 import type { FrontmatterHeading } from 'src/types/frontmatter'
 import TocNav from './toc-nav'
+import {useRouter} from "next/router";
 
 interface TableOfContentProps extends BoxProps {
   headings: FrontmatterHeading[]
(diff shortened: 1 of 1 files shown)


nikolovlazar:
@P0ppoff I think it'd be better to update the TOC based on the selected tab, instead of providing this fix. Can you come up with a solution for that? If not, let's leave the issue opened as a reminder.

I'll close this PR for the time being.

P0ppoff:
perfect, I have some free time Friday morning, I'll try it

nikolovlazar:
Awesome! The TOC is generated only from the `usage.mdx` file right now. We should have 3 TOCs generated at most, for the `usage.mdx`, `props.mdx`, and `theming.mdx`, and swap out the TOC data as the tabs selection change.

PR #1148, opened 2022-10-21
feat(docs):  MeteorJS in Framework guides section

Grubba27:
## 📝 Description

As for Meteor 2.8.0 we now have the option to create a new project using the Chakra-Ui, we in the core team personally love the ideals and the lib itself it would be an honor to be among the Frameworks guide

Added new section on framework guides with Meteor installation process

no


nikolovlazar (approved):
That's awesome! Thank you for providing a first-class support for Chakra UI, @Grubba27 🙌

PR #1165, opened 2022-10-25
docs: Update use-clipboard.mdx

jakegiganti:
Adds the setValue field and an example using the new setValue action

## 📝 Description

> Now that `useClipboard` supports `setValue` we should update the docs to reflect the new functionality. https://github.com/chakra-ui/chakra-ui/pull/6763

## 💣 Is this a breaking change (Yes/No):
No

PR #1175, opened 2022-10-29
feat: update `useClipboard` doc regarding the new `setValue` method

tuananhl:
## 📝 Description

- Update the `useClipboard` doc regarding the new method returned in this hook named `setValue`. 
- Add an example to help understand how `setValue` work.

No

### DIFF
--- a/content/docs/hooks/use-clipboard.mdx
+++ b/content/docs/hooks/use-clipboard.mdx
@@ -13,7 +13,7 @@ The `useClipboard` hook takes the following arguments:
 
 | Name               | Type                 | Required | Description                                                                                                                                                                      |
 | ------------------ | -------------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `text`             | `string`             | `true`   | The text or value that is to be copied.                                                                                                                                          |
+| `initialValue`     | `string`             | `true`   | The initial text that is to be copied.                                                                                                                                          |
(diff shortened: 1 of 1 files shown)


content/docs/hooks/use-clipboard.mdx
```diff
@@ -46,6 +47,39 @@ function Example() {
...
+    </>
+  )
+}
+```
+
+Using the `setValue` method to update the value to be copied.
+Remember, this method just changes the current value of hook that will be copied by calling `onCopy` method then.
+So, if we want to copy the new value into clipboard. we need to fire `onCopy` method again.
+
+Note: the value of `useClipboard` hook is not updated immediately after calling `setValue` method because it's a state. So, if we want to copy the new value into clipboard after we call `setValue` method, we should use `useEffect` hook to listen the change and trigger the `onCopy` method with new value.
```
nikolovlazar, line 67: Suggestion:

```
Use the `setValue` method to update the value that we want to copy. This method does not update the clipboard, it just changes the current value. So, if we want to copy the new value into clipboard we need to execute the `onCopy` method again.

> Note: the value of `useClipboard` hook is not updated immediately after calling `setValue` method because it's a 
> state variable.
```

content/docs/hooks/use-clipboard.mdx
```diff
@@ -37,7 +38,7 @@ import { useClipboard } from '@chakra-ui/react'
 ```jsx
 function Example() {
   const [value, setValue] = React.useState('Hello world')
```
nikolovlazar, line 40: Let's remove this line and obtain the `value` variable from the `useClipboard` below.

nikolovlazar (changes requested):

nikolovlazar:
Closed by #1165.

PR #1204, opened 2022-11-04
refactor(showcase): add new company

pehcst:
I'm adding a new company to the showcase page that uses chakra-ui


nikolovlazar:
Hi @pehcst. Thank you for opening this PR, but currently we don't accept community entries for the showcase page.

PR #1323, opened 2022-12-29
docs(theming): added popover theming docs

### LINKED ISSUE #1080: Theming: Popover
### Subject

Popover

### Description

Create a basic theming documentation for the `Popover` component that showcases how you can style it.

Theming file: [Popover theming file](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/popover/theming.mdx)

### Template Reference

For single part components, feel free to use the `Button` component theming doc as a template reference: [Button theming](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/button/theming.mdx).

For multipart components, feel free to use the `Input` component theming doc as a template reference: [Input theming](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/input/theming.mdx).

GKosheev:
Closes #1080 

## 📝 Description

> Adding ```Popover``` theming docs

## 📝 Additional Information

> SVG image needs to be inserted


GKosheev:
I think I forgot to add a showcase in theming.mdx file, I'm going to make another commit

nikolovlazar:
@GKosheev here's the gist for the Popover Anatomy illustration: https://gist.github.com/nikolovlazar/15314f3db08a525be7e8ce07b51248d2

content/docs/components/popover/theming.mdx
```diff
@@ -2,3 +2,174 @@
...
+- F: `arrow`
+- G: `closeButton`
+
+## Theming properties
+
+The properties that affect the theming of the `Popover` component are:
+
+- `arrowPadding`: The arrow padding of the popover. Defaults to `8`
+- `gutter`: The distance or margin between the reference and popper. Defautls to
+  `8`
```
nikolovlazar, line 29: In this section we want to notice the theming properties that can be overridden that are within the theme. They usually are `variant`, `size` etc... The `arrowPadding` and `gutter` are style props, so they don't belong here. The `Popover` component actually doesn't have any theming properties, it just has a `baseStyle` defined.

content/docs/components/popover/theming.mdx
```diff
@@ -2,3 +2,174 @@
...
+
+```jsx live=false
+import { popoverAnatomy as parts } from '@chakra-ui/anatomy'
+import { createMultiStyleConfigHelpers, defineStyle } from '@chakra-ui/react'
+const { defineMultiStyleConfig } =
+  createMultiStyleConfigHelpers(parts.keys)
+export const popoverTheme = defineMultiStyleConfig({
+  defaultProps: {
+    arrowPadding: 6,
+    gutter: 6,
```
nikolovlazar, line 148: The `defaultProps` can contain only these values: `variant`, `size`, and `colorScheme`. Putting the `arrowPadding` and `gutter` is not going to work here.

nikolovlazar (changes requested):

content/docs/components/popover/theming.mdx
GKosheev: @nikolovlazar, would it be better to delete this section or to explain that the ```Popover``` doesn't have any theming properties?

nikolovlazar: It's better to explain that there's no defaults to override, but you can always create your own.

GKosheev:
@nikolovlazar, I've read the docs and rewrote a few sections. 

I've been trying to find the definition of default props for a couple of the components, such as ```Button``` (specifically ```colorScheme``` prop), but I couldn't.  

The only thing I could find is the general definition of ```ThemingProps``` interface and the way some of the ```Button``` interfaces extend it, for example: 

```typescript
interface ButtonProps extends HTMLChakraProps<"button">, ButtonOptions, ThemingProps<"Button"> {
}
```

Where can I find default props definitions for each of the components? (specifically in code)

nikolovlazar:
@GKosheev you can check out the Theme Source for each component. Here's the Button theme source: https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/button.ts. If you scroll at the bottom you'll see the `defaultProps` definition.

GKosheev:
I think the ```Showcase``` section could have been done better. Let me think about how to make it prettier, and I'll make a commit later.

I'm thinking of at least adding a change mode icon, adjusting color for different modes, and perhaps, showing & styling some additional components.

aacevski (changes requested):
I would prefer if we improve the showcase area right now, I would take some inspiration from sites like Dribble:
https://dribbble.com/tags/popover

If you are still struggling, I can help you.

aacevski (changes requested):
Hey, I'm not that satisfied with the styling, we can work on them a little, here's some specific examples I found:
https://dribbble.com/shots/20392799-Notifications-Tetrisly-Design-System
https://dribbble.com/shots/19771053-Onboarding-modals-Untitled-UI
https://dribbble.com/shots/18567069-Accounts-modal-Untitled-UI

GKosheev:
@aacevski, let me know if that works. I couldn't think of anything regarding color variants, so I made a `rounded` version instead.

I liked the notifications example, so I completely followed it.

aacevski:
This looks better, can you just fix the conflicts?

GKosheev:
@aacevski, I recently found a way to change the value based on the color mode using `useColorModeValue` hook, so there is no need of passing `colorMode` to the other components. I can refactor the code slightly, so it looks cleaner.

aacevski (approved):
Hey, sorry for the late response. This looks good and I will merge it. Thanks again! 💚

PR #1337, opened 2023-01-03
docs(theming): added skeleton theming docs

### LINKED ISSUE #1085: Theming: Skeleton
### Subject

Skeleton

### Description

Create a basic theming documentation for the `Skeleton` component that showcases how you can style it.

Theming file: [Skeleton theming file](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/skeleton/theming.mdx)

### Template Reference

For single part components, feel free to use the `Button` component theming doc as a template reference: [Button theming](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/button/theming.mdx).

For multipart components, feel free to use the `Input` component theming doc as a template reference: [Input theming](https://github.com/chakra-ui/chakra-ui-docs/blob/main/content/docs/components/input/theming.mdx).

GKosheev:
Closes #1085 

## 📝 Description

> Adding ```Skeleton``` theming docs


configs/sandpack-contents/component-theming/skeleton.js
```diff
@@ -0,0 +1,117 @@
...
+              variant="red"
+            >
+              <Text>Hidden Text</Text>
+            </Skeleton>
+          </Box>
+  
+          <Box>
+            <Button colorScheme={"blue"} onClick={() => setLoading(!loading)}>
+              Toggle
+            </Button>
```
aacevski, line 50: Let's name this:
Toggle loading state

And if you can restyle the button since it doesn't go well with the skeletons.

configs/sandpack-contents/component-theming/skeleton.js
```diff
@@ -0,0 +1,117 @@
...
+          <Box>
+            <Text>With red variant</Text>
+            <Skeleton
+              isLoaded={!loading}
+              bg="orange"
+              fadeDuration={1}
+              variant="red"
+            >
+              <Text>Hidden Text</Text>
+            </Skeleton>
```
aacevski, line 44: Can we make the skeletons a little bit smaller when it comes to the width?

aacevski: On top of that, can we make one skeleton with different fadeDuration than 1?

configs/sandpack-contents/component-theming/skeleton.js
```diff
@@ -0,0 +1,117 @@
...
+    Skeleton,
+    Button,
+    IconButton,
+    useColorMode,
+  } from "@chakra-ui/react";
+  import { FaMoon, FaSun } from "react-icons/fa";
+  import { useState } from "react";
+  
+  export default function App() {
+    const [loading, setLoading] = useState(true);
```
aacevski, line 15: Let's rename these to isLoading, setIsLoading since generally booleans should have `is` prefix.

aacevski (changes requested):

configs/sandpack-contents/component-theming/skeleton.js
GKosheev: @aacevski, I've changed ```fadeDuration``` to 1, 1.5, and 2 for each of the skeletons respectively.

aacevski (changes requested):

aacevski (changes requested):
Can you please fix the conflicts and one more thing that is missing is styling for the Circle and Text skeletons since we mention them on the Usage tab.
image

GKosheev:
@aacevski, it seems to me that ```SkeletonText``` behaves slightly different from ```Skeleton``` and ```SkeletonCircle```. I was not able to achieve any color changes by giving it the same ```red``` variant as I gave to other components.

I found an interface and was able to identify all the props which I can change within ```SkeletonText``` itself. 

```typescript
interface SkeletonTextProps extends SkeletonProps {
    spacing?: SkeletonProps["margin"];
    skeletonHeight?: SkeletonProps["height"];
    startColor?: SkeletonProps["startColor"];
    endColor?: SkeletonProps["endColor"];
    isLoaded?: SkeletonProps["isLoaded"];
}
```

When initialized, the skeleton text within a component is being wrapped with ```class="chakra-skeleton__group"``` class, and the ```red``` variant's ```startColor``` and ```endColor```, which I'm trying to use on that component, are not being passed to its children.

image



I also found that for this specific component, animation from ```startColor``` and ```endColor``` doesn't work at all, though if you pass those 2 props to the component, it passes those values to its children. 

Is everything that I mentioned above expected behavior? If so, how do I separately change the styles of ```SkeletonText``` and ```SkeletonCircle```?

aacevski:
Hmm you are right, the styles of the SkeletonCircle / SkeletonText get applied from the Skeleton. That's a nice discovery, I will bring it up to the team. Anyhow, all you need to do is resolve the conflicts and we should be good.

aacevski (approved):

GKosheev:
@aacevski if that's somewhat trivial to fix, I'd be happy to investigate with further contribution

GKosheev:
@aacevski, I think I found the problem.

The problem is that somehow `SkeletonText` doesn't receive default properties for `speed` and `fadeDuration` from `Skeleton` components. Everything works If I put back the `defaultProps` in `skeleton/src/skeleton.tsx` folder as so: 
```typescript
Skeleton.defaultProps = {
  fadeDuration: 0.4,
  speed: 0.8,
}
``` 
(I know that this is not the solution as it was deleted on purpose).

I also found that both `SkeletonCircle` and `SkeletonText` receive `undefined` for both `fadeDuration` and `speed`:

`skeleton-circle.tsx`:
```typescript
export const SkeletonCircle: React.FC<SkeletonProps> = ({
  size = "2rem",
  ...rest
}) => <Skeleton borderRadius="full" boxSize={size} {...rest} />
// rest.speed - undefined
// rest.fadeDuration - undefined
```
 
`skeleton-text.tsx`:

```typescript
export const SkeletonText: React.FC<SkeletonTextProps> = (props) => {
  const {
    noOfLines = defaultNoOfLines,
    spacing = "0.5rem",
    skeletonHeight = "0.5rem",
    className,
    startColor,
    endColor,
    isLoaded,
    fadeDuration, // undefined
    speed, // undefined
    children,
    ...rest
  } = props
...

```


One of the ways I found to fix that is to set default values of those 2 props while destructing values from `omitThemingProps(skeletonProps)`:

```typescript
const {
  ...
  fadeDuration = 0.4,
  speed = 0.8,
  ...rest
} = omitThemingProps(skeletonProps)
```

Unfortunately, I couldn't identify the reason why those default values are getting lost.

aacevski:
I'll open a seperate PR or create an issue if I come to some kind of a clean solution. Thank you for your investigation. Can you fix the conflicts one more time? Sorry.

GKosheev:
@aacevski, I fixed the conflicts. 

I'll also try to dive deep into this problem as I'm curious about a possible solution. I'll let you know if I find anything.

PR #1355, opened 2023-01-10
docs: Update linkoverlay usage with routing library

arjunlubana:
# Update the usage of Link Overlay with a routing library

## 📝 Description

> This PR updates the code demo usage of Link Overlay with a routing library or a framework that leverages a client routing library. The context of the code demo in the docs is NextJS using NextLink.

> When trying to hoist a link in an encapsulating component. The code demo provided by the docs results in a [Hydration Error](https://nextjs.org/docs/messages/react-hydration-error). It is an issue that is closed in the core Chakra UI library [here](https://github.com/chakra-ui/chakra-ui/issues/7090).

If this is done the docs way, the result wraps an <a> tag inside another <a> tag that results in the above hydration error. 
image

## 🚀 New behavior

> 
```
 <LinkBox as="article">
      <h2>
        <LinkOverlay as={NextLink} href="#" passHref>
          Some blog post
        </LinkOverlay>
      </h2>
      <p>
        As a side note, using quotation marks around an attribute value is
        required only if this value is not a valid identifier.
      </p>
      <a href="#">Some inner link</a>
    </LinkBox>
```
This would result in the following required output.

image

No


aacevski (approved):

aacevski:
Hey @arjunlubana thank you for taking care of this 💖 Can you fix the conflicts, it's good to be merged!

Yuripetusko:
This doesn't work well when href value is an URL object, in this case typescript complains

image

I think this should be clarified then in docs

aacevski:
> This doesn't work well when href value is an URL object, in this case typescript complains
> 
> image
> 
> I think this should be clarified then in docs

I completely agree, if you are still wondering how to do this, I would go for this:
```js
import { chakra, LinkOverlay as ChakraLinkOverlay } from "@chakra-ui/react";
import NextLink from 'next/link';

const MagicLinkOverlay = chakra(ChakraLinkOverlay, {
  shouldForwardProp: (prop) =>
    [
      'href',
    ].includes(prop),
});


const LinkOverlay = () => (
  <MagicLinkOverlay as={NextLink} p={4} passHref href={{
    pathname: '/',
  }}>
    Box
  </MagicLinkOverlay>
)

```

jonahallibone:
> I completely agree, if you are still wondering how to do this, I would go for this:

@aacevski This seems like something which should be fixed in the type system right? The `href` type should be merged when `as` is used and overload the prop from the `LinkOverlay` prop. If I ignore the ts lint warning it still functions as expected

cschroeter:
Thanks for taking your time to create this PR. The issue has already been resolved.

PR #1386, opened 2023-01-25
docs: fix typo for the border radius in the styled-system

MaeWolff:
## 📝 Description

Fixed typo on the styled-system folder for the border-radius docs

No


aacevski:
This is not a type, the plural of border radius is radii. If you change the theme to use radius instead of radii, it won't work. I will close this however thank you for contributing. 💖

PR #1390, opened 2023-01-26
Update nextjs-guide.mdx

timreach:
Import theme from _app, where it is defined. Docs currently suggest importing from a theme file which does not exist. Either demonstrate the creation of this file or apply these suggested small changes.

Closes # N/A

## 📝 Description

Docs fix: A small update to the create-next-app guide for clarity. Currently it is referencing a file that would not exist for a user which could be confusing for someone new to react/next

Nothing, only change to docs

## 🚀 New behavior

Now allows user to follow along without issue

No

### DIFF
--- a/content/getting-started/nextjs-guide.mdx
+++ b/content/getting-started/nextjs-guide.mdx
@@ -172,6 +172,7 @@ function MyApp({ Component, pageProps }) {
 }
 
 export default MyApp
+export {theme}
 ```
 
 > To further customize components or build your own design system, the
@@ -193,7 +194,7 @@ for local storage syncing to work correctly.
 
 import { ColorModeScript } from '@chakra-ui/react'
 import { Html, Head, Main, NextScript } from 'next/document'
-import theme from './theme'
+import theme from './_app'
 
 export default function Document() {
   return (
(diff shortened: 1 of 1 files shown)


content/getting-started/nextjs-guide.mdx
```diff
@@ -110,7 +111,7 @@ for local storage syncing to work correctly.
 
 import { ColorModeScript } from '@chakra-ui/react'
 import { Html, Head, Main, NextScript } from 'next/document'
-import theme from './theme'
+import theme from './_app'
```
aacevski, line 114:
```suggestion
import { theme } from './_app'
```

content/getting-started/nextjs-guide.mdx
```diff
@@ -89,6 +89,7 @@ function MyApp({ Component, pageProps }) {
 }
 
 export default MyApp
+export {theme}
```
aacevski, line 92: Remove this and remove part and do export right where the const of the theme is defined.

export theme = extendTheme(...)

aacevski (changes requested):
Hey, thank you for your contribution 💖 Some small changes and we should be good!

noobinthisgame (approved):
Looks like this ready 😄

PR #1421, opened 2023-02-11
Updated package version to v1 in NextJS getting started docs

### LINKED ISSUE #1402: Fix version to v1 in the install command of v1 Nextjs docs
### Subject

Getting Started with Nextjs

### Description

I wanted to use Chakra UI in a Next.js project with React 17 so I ended up here:
https://v1.chakra-ui.com/guides/getting-started/nextjs-guide

The install command mentioned is:

```sh
npm i @chakra-ui/react @emotion/react@^11 @emotion/styled@^11 framer-motion@^6
```

This installs Chakra UI v2. I figured out after I started getting weird errors and noticed that I had Chakra v2 installed.

The command should be changed to use `@chakra-ui/react@^1` so it installs v1.

I'd be happy to make the change but I can't find the source code of v1 docs.

enzofalone:
Closes #1402

## 📝 Description

Both npm and yarn commands install the v1 version of chakra-ui in the NextJS docs

Only changes to v1 docs

## 🚀 New behavior

Now the package installed for ```@chakra-ui/react``` is up to the version the docs are written for (v1) for users that are getting started with the v1 version

No

### DIFF
--- a/pages/guides/getting-started/nextjs-guide.mdx
+++ b/pages/guides/getting-started/nextjs-guide.mdx
@@ -27,11 +27,11 @@ author: estheragbaje
 In your Next.js project, install Chakra UI by running either of the following:
 
 ```bash
-npm i @chakra-ui/react @emotion/react@^11 @emotion/styled@^11 framer-motion@^6
+npm i @chakra-ui/react@^1 @emotion/react@^11 @emotion/styled@^11 framer-motion@^6
 ```
 
 ```bash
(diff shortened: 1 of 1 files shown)


enzofalone:
Other sections such as the getting started guide for ```Create React App```, ```Vite```, and ```Remix``` should also be updated so the version installed is the v1 version.

noobinthisgame (approved):
Good catch! Thanks :)

PR #1470, opened 2023-03-12
Image with visible border example

carvalhochris:
Closes # 

## 📝 Description

Adding an example with JSX on how to make a profile image with a visible border

Non-existing example

## 🚀 New behavior

Adds another section to the page with the example

Not a breaking change


noobinthisgame:
Hi thanks for opening this PR. In my opinion adding this doesn't provide much more informations than the example above except some styling which isn't something specific to the image component. Closing this for now.

PR #1537, opened 2023-04-20
docs(RadioButton docs): fixes to custom Radio Button ReadMe example code

Julchu:
## 📝 Description

Documentation for Radio Button had example code that was throwing error for wrong variables

Property `getRadioProps` does not exist in `UseRadioProps`, fixed with getCheckboxProps since variable being assigned is `const checkbox`

Likewise, property `children` does not exist in `UseRadioProps`, fixed with `value` to show the value of each option

Self-close RadioCard element since RadioCard doesn't have a children prop

Documentation

## 🚀 New behavior

Nothing, just documentation fix (example code will now stop displaying error)

## 📝 Additional Information

### DIFF
--- a/content/docs/components/radio/usage.mdx
+++ b/content/docs/components/radio/usage.mdx
@@ -122,10 +122,10 @@ hooks to see more detail about their uses.
 ```jsx manual=true
 // 1. Create a component that consumes the `useRadio` hook
 function RadioCard(props) {
-  const { getInputProps, getRadioProps } = useRadio(props)
+  const { getInputProps, getCheckboxProps } = useRadio(props)
 
   const input = getInputProps()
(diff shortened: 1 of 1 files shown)


Riley1101:
Hello thanks for finding this out , I believe getCheckboxProps is now deprecated in the latest version and `variable being assigned is const checkbox` actually should have been something like `const radio` in the current docs.  I believe `<RadioCard/>` also need to render its children components.

noobinthisgame:
As correctly stated by @Riley1101 getCheckboxProps is deprecated and replaced by getRadioProps, see https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/radio/src/use-radio.ts#L303

PR #1553, opened 2023-04-30
refactor: improvement Table of Content usage in Usage Tab

iyu88:
## 📝 Description

- Improve the discomfort felt in the Table of Contents (hereinafter referred to as "TOC") on the right side while using Chakra-ui-docs.

- There is a problem that the selected text is not accurately highlighted (being bold) when you click the menu in TOC.
- There is also a problem that \<h2>Usage\</h2> does not come up to the top normally when the 'Usage' menu is pressed.

## 🚀 New behavior

- When you click the TOC menu, the menu you click is highlighted normally and works as intended.
- No matter which TOC menu is pressed, the scroll is moved so that the corresponding h-tag is positioned top.

- No

## 📝 Additional Information

Currently, TOC does NOT work when user select Props or Theming tab.
Because even user select other tabs, Usage container still remains on DOM structure and interferes with the use of TOC.
I think we need a new issue for that function.


src/components/page-container.tsx
```diff
@@ -92,7 +92,7 @@ function PageContainer(props: PageContainerProps) {
                       </Badge>
                     )}
                     {children}
-                    <Box mt='40px'>
+                    <Box mt='40vh'>
```
noobinthisgame, line 95: This change causes every page to have a huge whitespace at the bottom. This should not happen.

PR #1565, opened 2023-05-09
Update usage.mdx

joshmoy:
Closes # 

## 📝 Description

> Add `AlertDialogCloseButton` to the import statement

> When copying the code snippet on the `Alert Dialog` component code, the file breaks because `AlertDialogCloseButton` is not part of the imported components

## 🚀 New behavior

> This PR fixes the problem by adding `AlertDialogCloseButton` to the imports.


noobinthisgame (approved):
Good catch @joshmoy :)

PR #1573, opened 2023-05-11
docs: use correct props method

theoomoregbee:
## 📝 Description

Use the correct props-getter method.


noobinthisgame:
Hi @theoomoregbee,
as stated in the code of [`useRadio`](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/radio/src/use-radio.ts#L303) the `getCheckboxProps` function is deprecated for this hook and replace by `getRadioProps`.

Here 2 Sandboxes where it works:
https://codesandbox.io/s/flamboyant-maria-j0sppy
https://codesandbox.io/s/nice-benji-ir6ysk

PR #1611, opened 2023-06-30
docs(Card): Updated import path in docs.

bobba76:
**Updated import path.** 

Earlier import path wasn't working. Found this out when I tried to import using `'@chakra-ui/react'`and got: `Module '"@chakra-ui/react"' has no exported member 'Card'.`


segunadebayo:
This is incorrect. You can now import from @chakra-ui/react

PR #1612, opened 2023-07-01
docs/next13-app-directory

onurhan1337:
## 📝 Description

When we use Providers in the `<body>` tag, many components experience errors and problems. Until this situation reaches a better solution, it makes the most sense to keep the `body` tag inside the `children` in the `provider`.

> App directory content in Nextjs getting started page

## 🚀 New behavior

> Since I deleted the body tag that encapsulates the Provider, most of the problems with the components were solved for a while.

## 📝 Additional Information

Nextjs13 brought many innovations with it. This solution may not be very long term as it will take time to adapt to it. Nevertheless, we can prevent problems that may arise in the components of people who create a project from scratch (I encountered a few issues in the chakra-ui repository about this, 
an example: [#7817](https://github.com/chakra-ui/chakra-ui/issues/7817#issue-1782118623)) .

### DIFF
--- a/content/getting-started/nextjs-guide.mdx
+++ b/content/getting-started/nextjs-guide.mdx
@@ -40,7 +40,7 @@ function MyApp({ Component, pageProps }) {
   )
 }
 
-export default MyApp
+export default MyApp;
 ```
 
 ### App Directory Setup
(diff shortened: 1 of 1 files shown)


content/getting-started/nextjs-guide.mdx
```diff
@@ -70,33 +69,39 @@ To use Chakra UI in app directory, you can wrap `ChakraProvider` and
...
   return (
-    <html lang='en'>
-      <Providers>{children}</Providers>
+    <html lang="en">
+        <Providers>
+          {children}
+        </Providers>
     </html>
-  )
+  );
```
onurhan1337: Only changed there. Deleted `<body>` tag.

PR #1622, opened 2023-07-18
docs(Toast): Clarified when to use ToastContainer with createStandaloneToast

andrew8088:
Resolves confusion found in #7547

## 📝 Description
Currently, the docs don't accurately describe how to use `ToastContainer`: it sounds like you need to include `ToastContainer` in order to create toasts from outside React, but really, you only need to include it to create toasts outside of `ChakraProvider`.

Docs for `createStandaloneToast` are incomplete.

## 🚀 New behavior
Updated the wording of the `createStandaloneToast` docs.

No


cschroeter:
@andrew8088 

Thanks mate.

Issue #1645, opened 2023-08-29
Sub link navigation not working


imMatheus:
### Description

On all doc pages for the components (from what I see), in the **Props** and **Theming** pages, the links on the right side are not working. They are suppose to scroll that segment into view, but nothing happens. 

I expect to be able to go the [Menu Props page](https://chakra-ui.com/docs/components/menu/props), then in the right sidebar, press a link like [MenuList Props](https://chakra-ui.com/docs/components/menu/props#menulist-props) and then have the website scroll to that part. This does not happen, al

https://github.com/chakra-ui/chakra-ui-docs/assets/77362975/793cdc6b-0cb7-41af-a80b-c1e57f1437fe

though the URL changes which is something at least. But re-loading the website still does not scroll the right section into view

imMatheus:
For context, i am using a MacBook Pro, M2. On Google chrome

nahomtefera:
I submitted a pr to fix this issue

PR #1658, opened 2023-09-25
updating component.tsx to fix issue with anchor sub links not working…

nahomtefera:
#1645 … on Props and Theming tabs

https://github.com/chakra-ui/chakra-ui-docs/issues/1645

## 📝 Description
The anchor links on the "Props" and "Theming" tabs aren't working properly in the "On This Page" side bar.

image

> Add a brief description

When clicking on the anchor links on the right sidebar in the Props and Theming tabs, the view doesn't scroll into the correct section.

After digging into this, here's what I found:

When switching to Props and Theming tabs in the doc, the inner html of the Usage tab gets updated with the data for the Props or Theming tabs 

The problem lies in the Component.tsx file. There's a condition in the map iteration that displays children when index === 0 (which is the "Usage" tab), causing unexpected behavior.

      {tabsData.map((item, index) => (
        <Box key={index} id={item.id} hidden={!tabsData[index].match}>

          {index === 0 ? children : <MDXContent doc={item.doc} />}

        </Box>
      ))}

## 🚀 New behavior

To fix this, I've modified the code to remove the MDXContent component from the conditional. This ensures that the "Usage" Tab only receives the doc that corresponds to it.

      {tabsData.map((item, index) => (
        <Box key={index} id={item.id} hidden={!tabsData[index].match}>

          {index === 0 ?? children}

          <MDXContent doc={item.doc} />

        </Box>
      ))}
NO

### DIFF
--- a/configs/search-meta.json
+++ b/configs/search-meta.json
@@ -48273,6 +48273,761 @@
     }
   },
   {
+    "content": "Version 2.8.1",
+    "id": "//changelog/v2.8.1",
+    "type": "lvl1",
+    "url": "/changelog/v2.8.1",
+    "hierarchy": { "lvl1": "Version 2.8.1" }
+  },
+  {
+    "content": "@chakra-ui/tabs@3.0.0",
+    "id": "//changelog/v2.8.1",
+    "type": "lvl2",
+    "url": "/changelog/v2.8.1#chakra-uitabs300",
+    "hierarchy": {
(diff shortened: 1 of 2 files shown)


cschroeter:
Thanks mate but there are still some issues with your suggested solution. I will have a closer look once we moved the docs repo into the chakra ui mono repo.

PR #1662, opened 2023-10-04
feat: add bun package manager commands

jakeboone02:
## 📝 Description

Adds [Bun](https://bun.sh/) commands to the docs, beside existing `npm`/`yarn`/`pnpm` commands.

Docs do not have Bun commands

## 🚀 New behavior

Docs now have Bun commands.

No.

## 📝 Additional Information

I added the Bun logo to the codebase because the `react-icons` package doesn't include any icon sets that have the Bun logo yet. Also it's in full color rather than monochrome like the others because frankly it looks kind of creepy in light mode (dark logo on white background) with those empty eye sockets:

image vs image


cschroeter:
@jakeboone02 

Thanks mate

PR #1696, opened 2024-01-16
Update nextjs-app-guide.mdx to add the cssVarsRoot to the provider

Emyboy:
Without it there'll be a typescript error in my NextJs layout.tsx in my app directory:

Property 'cssVarsRoot' is missing in type '{ children: ReactNode; }' but required in type 'ChakraProviderProps'

and adding cssVarsRoot to the provider fixed the issue.


cschroeter:
@Emyboy 

Thanks for taking your time to create this PR. I've tried this myself but could not find an issue with this `cssVarsRoot` also in the types it is marked as optional:

```ts
interface ThemeProviderProps extends ThemeProviderProps$1 {
    cssVarsRoot?: string;
}
```

So I assume the error is somewhere else.

